PR #1, opened 2021-11-18
Add README

### LINKED ISSUE #2: document sections and tags
after reading about sections and tags, we should have a small paragraph in our documentation explaining what those are, especially in the context of this library

nforro:
Fixes #2

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,171 @@
+# specfile
+
+Python library for parsing and manipulating RPM spec files. Main focus is on modifying existing spec files, any change should result in a minimal diff.
+
+This project is still a work in progress.
+
+## Motivation
+
+Originally, [rebase-helper](https://github.com/rebase-helper/rebase-helper/) provided an API for spec file modifications that was also used by [packit](https://github.com/packit/packit). The goal of this project is to make the interface more general and convenient to use by not only packit but also by other Python projects that need to interact with RPM spec files.
+
+## Important terms used in this library
+
+### Section
+
+Section is a spec file section, it has a well-defined name that starts with _%_ character and that can optionally be followed by arguments.
+
+In this library, the starting _%_ of section name is ommited for convenience.
+
+There is a special section internally called `%package`, often also referred to as preamble, and it represents the content of the spec file that preceeds the first named section (usually `%description`). This section contains the main package metadata (tags). Metadata of subpackages are defined in subsequent `%package` sections, that are not anonymous and are always followed by arguments specifying the name of the subpackage (e.g. `%package doc` or `%package -n completely-different-subpackage-name`).
+
+### Tag
+
+Tag represents a single item of metadata of a package. It has a well-defined name and a value. Tags are defined in `%package` sections.
+
+For the purposes of this library, a tag can have associated comments. These are comment lines that are in a spec file directly above the tag definition.
+
+## Examples and use cases
+
+The following examples should cover use cases required by [packit](https://github.com/packit/research/blob/main/specfile/README.md).
+
+### Instantiating
+
+```python
+from specfile import Specfile
+
+# using an absolute path
+specfile = Specfile('/tmp/test.spec')
+
+# using a relative path and a different sourcedir
+specfile = Specfile('test.spec', sourcedir='/tmp/sources')
+```
+
+### Reloading
+
+```python
+# if the spec file happens to be modified externally, it can be reloaded
+specfile.reload()
+```
+
+### Saving changes
+
+```python
+# no autosave
+specfile = Specfile('test.spec')
+...
+# saving explicitly when needed
+specfile.save()
+
+# enabling autosave, changes are saved immediately after any modification
+specfile = Specfile('test.spec', autosave=True)
+
+# as a context manager, saving is performed at context exit
+with Specfile('test.spec') as specfile:
+    ...
+```
+
+### Low-level manipulation
+
+```python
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -0,0 +1,117 @@
...
+
+specfile uses RPM for parsing spec files and macro expansion. Unfortunately, macros are always stored in a global context, which poses a problem for multiple instances of Specfile.
+
+## Examples and use cases
+
+### Instantiating
+
+```python
+# using an absolute path
+specfile = Specfile('/tmp/test.spec')
```
FrNecas, line 17: Hm, I think we should unify the naming a bit. The introduction paragraph uses `spec file` (note the space), whereas the name of the class is `Specfile`. We should either use `RPM specfile` above, or name the class `SpecFile` in my opinion

nforro: `RPM spec file` IMO equals to `RPM .spec file`, and it should stay that way, so it is clear even to those who have no idea what's going on. I thought about `SpecFile` at first, but then I decided to go with the single-word variant (which is being used quite often I think), because it's more straightforward, you can write `specfile` and not `spec_file` etc.

Also, packit uses Specfile as well:
https://github.com/packit/packit/blob/4f7a3690c68c1bf730678e92cc648d8610c850dd/packit/specfile.py#L25

README.md
```diff
@@ -0,0 +1,117 @@
+# specfile
+
+Python library for parsing and manipulating RPM spec files. Main focus is on modifying existing spec files, any change should result in a minimal diff.
```
FrNecas, line 3: I'd add a sentence or two about motivation. Something like `Previously, [rebase-helper](https://github.com/rebase-helper/rebase-helper/) provided an API for the modifications which was also used by [packit](https://github.com/packit/packit). The goal is to make the interface more general and convenient to use by these projects but also by other python projects which need to interact with RPM spec files.`

README.md
```diff
@@ -0,0 +1,117 @@
...
+
+## High-level
+
+### Version and release
+
+```python
+# getting version and release
+print(specfile.version)
+print(specfile.release)
+# setting version and release
```
FrNecas, line 82: I'd mention that this sets it as raw content. On one hand, this is not official documentation, however we should make the expectations clear even here.

README.md
```diff
@@ -0,0 +1,117 @@
...
+
+Python library for parsing and manipulating RPM spec files. Main focus is on modifying existing spec files, any change should result in a minimal diff.
+
+## Caveats
+
+### RPM macros
+
+specfile uses RPM for parsing spec files and macro expansion. Unfortunately, macros are always stored in a global context, which poses a problem for multiple instances of Specfile.
+
+## Examples and use cases
```
FrNecas, line 11: We should maybe mention for now that most of these have not been implemented yet and are just examples of the intended future interface

README.md
```diff
@@ -0,0 +1,117 @@
...
+# setting both at the same time (release defaults to 1)
+specfile.set_version_and_release('2.1', release='3')
+# setting version while trying to preserve macros
+specfile.set_version_and_release('2.1', preserve_macros=True)
+```
+
+### Changelog
+
+```python
+specfile.add_changelog_entry('New upstream release 2.1')
```
FrNecas, line 94: Perhaps a comment mentioning that the author/email are taken from git config could be useful

README.md
FrNecas: Also maybe a sentence like `The following examples should cover use cases described by [packit](https://github.com/packit/research/blob/main/specfile/README.md)`

lachmanfrantisek (commented):
The code snippets look really good. Thanks!

README.md
nforro: Ok, I'd just add it as another paragraph.

README.md
nforro: I agree, but then it's necessary to explain what raw means in this context.

README.md
nforro: packit actually asks for what rpmdev-bumpspec is using, which is the output of [rpmdev-packager](https://pagure.io/rpmdevtools/blob/main/f/rpmdev-packager).

nforro:
@FrNecas I've tried to incorporate your suggestions and also made some small design changes.

README.md
TomasTomecek: this is a good point

on the other hand, we started unifying on "spec file" in our user documentation

+1 Specfile class name

README.md
```diff
@@ -0,0 +1,157 @@
+# specfile
+
+Python library for parsing and manipulating RPM spec files. Main focus is on modifying existing spec files, any change should result in a minimal diff.
+
+This project is still work in progress.
```
TomasTomecek, line 5:
```suggestion
This project is still a work in progress.
```

README.md
```diff
@@ -0,0 +1,157 @@
...
+### Saving changes
+
+```python
+# no autosave
+specfile = Specfile('test.spec')
+...
+# saving explicitly when needed
+specfile.save()
+
+# enabling autosave, changes are saved immediately after any modification
```
TomasTomecek, line 49: this is mind boggling, I love it!

TomasTomecek (approved):
👍

nforro:
I've made some more changes, @FrNecas could you take a look?

TomasTomecek (approved):

FrNecas (approved):
Looks great, thanks for the improvements, lets get to implementing it!

PR #3, opened 2022-01-11
Basic implementation

nforro:


specfile/tags.py
```diff
@@ -0,0 +1,446 @@
...
+        data: List of individual tags.
+    """
+
+    def __init__(
+        self, data: Optional[List[Tag]] = None, remainder: List[str] = []
+    ) -> None:
+        super().__init__()
+        if data is not None:
+            self.data = data.copy()
+        self._remainder = remainder.copy()
```
jpopelka, line 353: [Don't use `[]` as a default argument value (for `remainder`).](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)

```suggestion
        self, data: Optional[List[Tag]] = None, remainder: Optional[List[str]] = None
    ) -> None:
        super().__init__()
        if data is not None:
            self.data = data.copy()
        self._remainder = remainder.copy() if remainder is not None else []
```

specfile/specfile.py
```diff
@@ -0,0 +1,101 @@
...
+        return self
+
+    def __exit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_value: Optional[BaseException],
+        traceback: Optional[types.TracebackType],
+    ) -> None:
+        if not self.autosave:
+            self.save()
```
jpopelka, line 50: Is the condition OK? (should the `not` be there?)

specfile/tags.py
```diff
@@ -0,0 +1,445 @@
...
+        """Value of the tag after expanding macros and evaluating all conditions."""
+        return self._expanded_value
+
+    @property
+    def comments(self) -> List[str]:
+        """List of comment lines associated with the tag."""
+        comments = itertools.takewhile(
+            lambda x: x.startswith("#"), reversed(self._preceding_lines)
+        )
+        return [c[1:] for c in reversed(list(comments))]
```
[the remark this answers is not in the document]
nforro: Because the result is supposed to be a list of consecutive comment lines directly above the tag line, so it is necessary to iterate from the bottom up.

specfile/tags.py
```diff
@@ -0,0 +1,445 @@
...
+        try:
+            return next(
+                iter(
+                    i
+                    for i in range(len(self.data))
+                    if self.data[i].name.capitalize() == name.capitalize()
+                )
+            )
+        except StopIteration:
+            raise ValueError
```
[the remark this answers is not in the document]
nforro: No, it doesn't, let me fix that.

specfile/specfile.py
nforro: The point of the condition is not to unnecessarily write to the spec file multiple times (at this point it should have been already saved if a change was made), but perhaps it's better to save no matter what.

specfile/tags.py
nforro: Thanks, I'll fix this.

specfile/rpm.py
```diff
@@ -0,0 +1,253 @@
...
+class RPM:
+    @staticmethod
+    def parse(
+        s: str, sourcedir: Path, macros: Optional[List[Tuple[str, str]]] = None
+    ) -> rpm.spec:
+        """
+        Parses the content of a spec file.
+
+        Args:
+            s: String representing the content of a spec file.
```
FrNecas, line 232: Is there a reason behing this naming? I'd probably much rather see `content` rather than `s`.

specfile/rpm.py
```diff
@@ -0,0 +1,253 @@
...
+            sourcedir: Path to sources and patches.
+            macros: List of macro definitions.
```
FrNecas, line 234: Maybe a note that this influences the parsing could be useful? Something like `List of macro definitions which will be applied before parsing`

specfile/specfile.py
```diff
@@ -0,0 +1,106 @@
...
+from typing import Iterator, List, Optional, Tuple, Type, Union
+
+from specfile.rpm import RPM, Macros
+from specfile.sections import Sections
+from specfile.tags import Tags
+
+
+class Specfile:
+    """
+    Class representing spec file.
```
FrNecas, line 16: Nitpick: missing article for `spec file` (slighly lower `the` is used but `a` is probably more suitable here)

specfile/specfile.py
```diff
@@ -0,0 +1,106 @@
...
+
+    Attributes:
+        path: Path to the spec file.
+        sourcedir: Path to sources and patches.
+        autosave: Whether to automatically save any changes made.
+        macros: List of macro definitions.
```
FrNecas, line 22: Suggestion: `extra macro definitions` to stress that this is something else than the macros defined in the specfile.

specfile/specfile.py
```diff
@@ -0,0 +1,106 @@
...
+        self,
+        path: Union[Path, str],
+        sourcedir: Optional[Union[Path, str]] = None,
+        autosave: bool = False,
+        macros: Optional[List[Tuple[str, str]]] = None,
+    ) -> None:
+        self.path = Path(path)
+        self.sourcedir = Path(sourcedir or self.path.parent)
+        self.autosave = autosave
+        self.macros = macros.copy() if macros is not None else None
```
FrNecas, line 35: Wouldn't an empty list `[]` be more convenient for further use than `None` if the argument is `None`?

specfile/tags.py
```diff
@@ -0,0 +1,439 @@
...
+        if not name or name.capitalize().rstrip("0123456789") not in TAG_NAMES:
+            raise ValueError(f"Invalid tag name: '{name}'")
+        self.name = name
+        self.value = value
+        self._expanded_value = expanded_value
+        self._separator = separator
+        self.comments, self._preceding_lines = self._extract_comments(preceding_lines)
+
+    def __repr__(self) -> str:
+        preceding_lines = repr(self._preceding_lines)
```
FrNecas, line 303: Hm, this doesn't include comments does it? As far as I understand, the input of constructor `preceding_lines` includes all lines before the tag, comments and possibly something "extra" (maybe macro definition?). As far as I can see, this makes this `repr` behave incorrectly, for example if `preceding_lines=['%define macro macro', '# comment']` then the `_preceding_lines` attribute becomes only `['%define macro macro']` and hence the `repr` magic method doesn't actually return the correct way of constructing the object. Am I missing something?

specfile/tags.py
```diff
@@ -0,0 +1,439 @@
...
+
+    @property
+    def expanded_value(self) -> str:
+        """Value of the tag after expanding macros and evaluating all conditions."""
+        return self._expanded_value
+
+    @staticmethod
+    def _extract_comments(lines: List[str]) -> Tuple[List[str], List[str]]:
+        comments = itertools.takewhile(lambda x: x.startswith("#"), reversed(lines))
+        rest = itertools.dropwhile(lambda x: x.startswith("#"), reversed(lines))
```
FrNecas, line 322: This should also probably consider whitespaces before comments, I suggest using a regex (`'  # comment'` should still be valid, I think). Or are whitespaces stripped somewhere else?

specfile/tags.py
```diff
@@ -0,0 +1,439 @@
...
+
+import collections
+import itertools
+import re
+from typing import List, Optional, Tuple
+
+from specfile.sections import Section
+
+# valid tag names extracted from lib/rpmtag.h in RPM source
+TAG_NAMES = [
```
FrNecas, line 12: This should probably be a set (for more efficient check of presence) since we are only checking for presence or iterating over it where order is not relevant.

specfile/tags.py
```diff
@@ -0,0 +1,439 @@
...
+        return f"Tags({data}, {remainder})"
+
+    def __getattr__(self, name: str) -> Tag:
+        try:
+            return self.data[self.find(name)]
+        except ValueError:
+            raise AttributeError(name)
+
+    def __setattr__(self, name: str, value: str) -> None:
+        if name.capitalize().rstrip("0123456789") not in TAG_NAMES:
```
FrNecas, line 366: Why is there `not in TAG_NAMES`? Shouldn't it be the other way around? Also a minor thing that this lets through tags such as `Arch0`. The number stripping should probably only be done for Patch and Source tags as far as I am aware.

specfile/tags.py
```diff
@@ -0,0 +1,439 @@
...
+    # remove Source1 tag
+    del tags.source1
+    ```
+
+    Attributes:
+        data: List of individual tags.
+    """
+
+    def __init__(
+        self, data: Optional[List[Tag]] = None, remainder: Optional[List[str]] = None
```
FrNecas, line 347: Maybe some documentation about what `remainder` is would be nice.

specfile/tags.py
```diff
@@ -0,0 +1,439 @@
...
+            index = r"\d?" if tag in ["Source", "Patch"] else ""
+            return fr"^(?P<n>{name}{index})(?P<s>\s*:\s*)(?P<v>.+)"
+
+        tag_regexes = [re.compile(regex_pattern(t), re.IGNORECASE) for t in TAG_NAMES]
+        data = []
+        buffer: List[str] = []
+        for line in raw_section:
+            m = next((m for m in (r.match(line) for r in tag_regexes) if m), None)
+            if m:
+                e = next(
```
FrNecas, line 411: I find this method a bit hard to read/understand at first sight (took me some time to understand what's going on). Maybe some comments could be helpful here, especially due to the use of short names (which is understandable), maybe a quick comment that we need to parse the tag in both the raw and expanded section.

FrNecas (changes requested):
Overall looks good to me, I have just a few questions and remarks.

specfile/specfile.py
nforro: I was trying to be consistent with other `packit` projects, namely `ogr`, there are no articles in class docstrings there. But it probably wasn't the best idea, I'll add the missing articles everywhere.

specfile/rpm.py
nforro: Ok, I'll change it.

specfile/rpm.py
nforro: I assumed it's obvious, but you are right, it's better to be explicit.

specfile/specfile.py
nforro: I don't think there will be any modification to the macros list before passing it to `RPM.parse()`, but sure, I'll change it to `[]` just in case.

specfile/tags.py
nforro: You are right, I'll fix this.

specfile/tags.py
nforro: Good point.

specfile/tags.py
nforro: It's an internal attribute, and nobody should be instantiating `Tags` directly, but I'll add some description for it.

specfile/tags.py
nforro: Fair enough, I'll add some comments.

specfile/tags.py
nforro: No, this ensures that "regular" attributes are processed normally (by calling `__setattr__()` of the parent). That's all. You are right about `Arch0`, but I didn't want to make the condition overly complex. Do you think it's necessary?

specfile/tags.py
nforro: Good catch, it used to include comments as well, but I changed that and forgot to account for it in `__repr__()`.

specfile/tags.py
nforro: Ok, thanks. The rest of the issues should be addressed.

nforro: > Ok, thanks. The rest of the issues should be addressed.

Actually, I just realized that `Tag.comments` and `Tag.__repr__()` is still not entirely correct, I'll fix it ASAP.

FrNecas (approved):
Looks good to me now, thanks for the changes. I will try to setup the CI in the upcoming days

nforro:
@FrNecas The comments turned out to be a much bigger deal than expected, so I reworked it substantially. It's not possible to use a simple list of strings, because the amount of whitespace preceding every comment must be preserved (in order to ensure minimal diff when making changes).
I've also fixed a bug in classes derived from `collections.UserList` not reimplementing `copy()`, leading to infinite recursion.

tests/unit/test_tags.py
```diff
@@ -0,0 +1,102 @@
...
+                ":    ",
+                Comments([], ["%global ver_major 1", "%global ver_minor 0", ""]),
+            ),
+            Tag("Version", "%{ver_major}.%{ver_minor}", "1.0", ": ", Comments()),
+            Tag("Release", "1%{?dist}", "1.fc35", ": ", Comments()),
+            Tag("Epoch", "1", "", ":   ", Comments([], ["", "%if 0"])),
+        ],
+        ["%endif"],
+    )
+    assert tags.reassemble() == [
```
FrNecas, line 91: Hm, would it be possible to add some comment lines to this test so that it also tests correct reassembly of comments, possibly with whitespaces?

specfile/tags.py
```diff
@@ -0,0 +1,599 @@
...
+    "Verifyscriptflags",
+    "Verifyscriptprog",
+    "Veritysignaturealgo",
+    "Veritysignatures",
+    "Version",
+    "Xpm",
+}
+
+
+class Comment:
```
FrNecas, line 275: I guess this could be implemented (and slightly simplified) using a [dataclass](https://docs.python.org/3/library/dataclasses.html) since it basically only stores the prefix and text which would remove the need for explicitly defining repr/str

FrNecas (approved):
The changes to comments seem nice, as far as I can tell, nice work. I don't see any more problems with the PR so we can basically merge, just two minor comments from me that could slightly simplify/improve the code.

tests/unit/test_tags.py
nforro: Good idea, will do.

specfile/tags.py
nforro: > which would remove the need for explicitly defining repr/str

For `__repr__()`, yes, but not for `__str__()`.

I'm not sure, is it worth it?

FrNecas: Good point, it probably wouldn't really make much difference, constructor and `__repr__` could be removed but that's about it and it wouldn't simplify it a lot since it's only 2 attributes. I am fine with keeping it the way it is now

tests/unit/test_tags.py
nforro: Done.

PR #4, opened 2022-01-21
Implement changelog parsing and manipulation

nforro:


nforro:
I'm going to create some test cases as well.

specfile/changelog.py
```diff
@@ -0,0 +1,127 @@
...
+    def __init__(
+        self, date: datetime.date, author: str, evr: str, content: List[str]
+    ) -> None:
+        self.date = date
+        self.author = author
+        self.evr = evr
+        self.content = content.copy()
+
+    def __str__(self) -> str:
+        return f"{self.header}\n" + "\n".join(self.content) + "\n"
```
FrNecas, line 57: Couldn't this type of joining cause some problems with extra blanks lines? I am not completely sure whether they are allowed in the changelog or not. (same for `Changelog.reassemble()`)

specfile/specfile.py
```diff
@@ -104,3 +113,66 @@ def tags(self, section: str = "package") -> Iterator[Tags]:
...
+
+    @contextlib.contextmanager
+    def changelog(self) -> Iterator[Optional[Changelog]]:
+        """
+        Context manager for accessing changelog.
+
+        Yields:
+            Spec file changelog as `Changelog` object or None if there is no %changelog section.
+        """
+        with self.sections() as sections:
```
FrNecas, line 125: Hm, this means that the changelog has to be parsed from scratch every time that the user of the library wants to change it, right? I had an idea for an alternative approach though I am not sure if it's really better. We could make `Changelog` inehrit from `Section` and it could have an `entries` attribute that would contain the list of entries similarly to the way it is now. This way, I guess it would be a bit easier to keep the two "views" on the changelog section (raw section and a list of changelog entries) in sync without having to parse it each time. What do you think about this approach?

specfile/specfile.py
```diff
@@ -104,3 +113,66 @@ def tags(self, section: str = "package") -> Iterator[Tags]:
...
+        with self.changelog() as changelog:
+            if not changelog:
+                return
+            evr = self.expand(
+                "%{?epoch:%{epoch}:}%{version}-%{release}", extra_macros=[("dist", "")]
+            )
+            changelog.insert(
+                0,
+                ChangelogEntry(
+                    date or datetime.date.today(), get_author(), evr, [f"- {entry}"]
```
FrNecas, line 176: This seems to assume that there is only a single "entry" in one changelog. But what if we wanted to create something like:
```
<header>
- something
- something else
```

README.md
```diff
@@ -104,6 +104,16 @@ with specfile.tags() as tags:
...
+
+# working with changelog
+with specfile.changelog() as changelog:
+    # most recent changelog entry
+    print(changelog[0])
+    # making changes
+    changelog[1].date = datetime.date.today()
+    changelog[1].content.append('- another line')
+    # removing the oldest entry
+    del changelog[-1]
```
lachmanfrantisek, line 116: I am wondering if using a reverse order wouldn't be more consistent/stable.

e.g.:
```python
>>> len(the_changelog)
3
>>> the_changelog[2]
v1.0 changelog entry
>>> the_changelog.add_changelog_entry()
>>> the _changelog[2]
v2.0 changelog entry
```

The old-to-new ordering will return the same changelog entries in this example. And using `[-1]` can still be nicely used to get the last one.

specfile/changelog.py
```diff
@@ -0,0 +1,127 @@
...
+    "Apr",
+    "May",
+    "Jun",
+    "Jul",
+    "Aug",
+    "Sep",
+    "Oct",
+    "Nov",
+    "Dec",
+]
```
lachmanfrantisek, line 34: Can't we use any library for this? I hope it's already defined somewhere.

specfile/changelog.py
```diff
@@ -0,0 +1,127 @@
...
+        date = repr(self.date)
+        content = repr(self.content)
+        return f"ChangelogEntry({date}, '{self.author}', '{self.evr}', {content})"
+
+    @property
+    def header(self) -> str:
+        """Entry header in the format that is being used in Fedora spec files."""
+        weekday = DAYS[self.date.weekday()]
+        month = MONTHS[self.date.month - 1]
+        return f"* {weekday} {month} {self.date:%d %Y} {self.author} - {self.evr}"
```
lachmanfrantisek, line 69: `rpmdev-bumpspec` uses time as well for me, e.g.:

```
 %changelog
* Mon Jan 24 14:13:29 CET 2022 Frantisek Lachman <flachman@redhat.com> - 0.5.3-1
- New upstream release 0.5.3
```

specfile/changelog.py
```diff
@@ -0,0 +1,127 @@
...
+            section: Section to parse.
+
+        Returns:
+            Constructed instance of `Changelog` class.
+        """
+        header_regex = re.compile(r"^\* \w+ (\w+) (\d{2}) (\d{4}) (.+) - (.+)$")
+        data = []
+        for line in section:
+            m = header_regex.match(line)
+            if m:
```
lachmanfrantisek, line 102: Can we use [walrus](https://www.python.org/dev/peps/pep-0572/) here? (What Python versions do we want to support?)

```suggestion
            if m := header_regex.match(line):
```

lachmanfrantisek (commented):
This looks really simple and clean, nice!

specfile/changelog.py
nforro: What extra blank lines do you mean? Could you give an example?

README.md
nforro: Makes sense, I'll change it.

specfile/changelog.py
nforro: I think we should support the same Python versions as packit. Are there plans to drop support for Python 3.6 and 3.7?

But nice idea, I didn't even think of it.

specfile/changelog.py
nforro: A library for this seemed like an overkill (I was thinking about [Babel](http://babel.pocoo.org/en/latest/), I'll try to find some lightweight alternative).

specfile/changelog.py
nforro: You got me there, I didn't realize RPM supports this (since [4.14.0](https://github.com/rpm-software-management/rpm/commit/57f94a582602f0353cdb17a02dc12c4461d4f32d)). So we should definitely default to that.

specfile/specfile.py
nforro: Thinking about it, I'd probably find more use cases for adding entries with mutiple lines myself. I'm going to change it.

specfile/changelog.py
FrNecas: Would be nice to use it but I recently used something specific from python 3.8 (where walrus operator was also added) and the CI was failing in EPEL.

specfile/changelog.py
FrNecas: What I thought about here was something like:
```
* Mon Jun 22 2022 Author theauthor@mail.com
<blank line here>
- entry
- entry
```

specfile/specfile.py
FrNecas: Makes sense, lets keep it the way as with `Tags`, thanks for the explanation

specfile/changelog.py
nforro: Yes, this is actually valid, RPM doesn't have any issues with it. However, this would happen only if user wanted to have blank lines there, am I right? In that case, why not just let him?

FrNecas: Oh, good point, for some reason I thought that the `__str__` method was being used for reassembly similar to sections. Looking at it again, the case mentioned above should hopefully work just fine as far as I can see (there will be a single changelog entry with `content` equal to `["", "- entry", "-entry"]`)

specfile/changelog.py
lachmanfrantisek: Good point with EPEL.

nforro:
I've addressed the comments, I decided to use [arrow](https://github.com/arrow-py/arrow) library for parsing and formatting timestamps.

I'm also going to write a test for `Specfile.add_changelog_entry()` (or should I do it later in another PR?).

specfile/changelog.py
```diff
@@ -0,0 +1,128 @@
...
+                        datetime = arrow.get(dt, "MMM DD YYYY").datetime
+                    except arrow.parser.ParserError as e:
+                        raise SpecfileException(
+                            "Failed to parse timestamp of changelog entry"
+                        ) from e
+                    else:
+                        legacy = True
+                else:
+                    legacy = False
+                data.insert(0, ChangelogEntry(datetime, author, evr, [], legacy))
```
lachmanfrantisek, line 108: Would it be possible to use keyword arguments to make this explicit? (I am speaking about the empty list.)
```suggestion
                data.insert(0, ChangelogEntry(datetime, author, evr, content=[], legacy=legacy))
```

specfile/specfile.py
```diff
@@ -104,3 +115,72 @@ def tags(self, section: str = "package") -> Iterator[Tags]:
...
+            if author is not None:
+                return author + (f" <{email}>" if email else "")
+            try:
+                result = subprocess.check_output("rpmdev-packager").decode().strip()
+            except subprocess.CalledProcessError as e:
+                raise SpecfileException("Failed to auto-detect author") from e
+            else:
+                return result
+
+        with self.changelog() as changelog:
```
lachmanfrantisek, line 171: :+1: for using a context manager --  both saving and manipulation looks really clear.

specfile/changelog.py
```diff
@@ -0,0 +1,128 @@
...
+            elif data and line:
+                data[0].content.append(line)
+        return Changelog(data)
+
+    def reassemble(self) -> List[str]:
```
lachmanfrantisek, line 113: I was a bit confused by the name.

What about `get_lines`/`get_raw`/`get_raw_sections`/`generate_text_repr`/`raw_content`/`raw_changelog`/`raw_section_lines` or anything else showing that we are exporting something and the result is in form of (raw) lines...

lachmanfrantisek (approved):
Looks really good to me!

Just two nit-picks...;)

specfile/changelog.py
nforro: I was trying to come up with a word that would be the opposite of "parse", but I'll replace it with something more clear.

lachmanfrantisek: Thanks.

specfile/changelog.py
```diff
@@ -0,0 +1,130 @@
...
+        Parses a %changelog section.
+
+        Args:
+            section: Section to parse.
+
+        Returns:
+            Constructed instance of `Changelog` class.
+        """
+        header_regex = re.compile(
+            r"^\* \w+ (\w+ \d{2} (?:\d{2}:\d{2}:\d{2} [A-Z]+ )?\d{4}) (.+) - (.+)$"
```
FrNecas, line 88: Nitpick: since this regex is getting a bit long (especially due to the date formats), it could be useful to use a verbose regex, would be a bit easier to understand. Or maybe at least add a comment with an example header line?

FrNecas (approved):
Looks clean to me, just one readability suggestion

specfile/changelog.py
nforro: I actually think using verbose regex would be less readable, I'll add example headers.

nforro: It turns out there are other header formats in use, for example [kernel](https://src.fedoraproject.org/rpms/kernel/blob/rawhide/f/kernel.spec#_3005) has an unusual way to separate evr. I think we need to support this as well.

specfile/changelog.py
```diff
@@ -0,0 +1,134 @@
...
+from typing import List, Optional, Union
+
+import arrow
+
+from specfile.sections import Section
+
+
+class ChangelogEntry:
+    """
+     Class that represents a changelog entry.
```
FrNecas, line 15: There is an extra space in the docstring if I see correctly

specfile/changelog.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def __repr__(self) -> str:
+        content = repr(self.content)
+        return f"ChangelogEntry('{self.header}', {content})"
+
+    @property
+    def extended_timestamp(self) -> bool:
+        """Whether the timestamp present in the entry header is extended (date and time)."""
+        try:
+            arrow.get(self.header, "ddd MMM DD YYYY")
```
FrNecas, line 41: Are you sure that this works correctly? Since you are passing the whole header to `arrow.get` now, this will always throw an exception, won't it? Since `self.header` is the whole `* <datetime> <author> <NVR>` string.

specfile/specfile.py
```diff
@@ -104,3 +115,68 @@ def tags(self, section: str = "package") -> Iterator[Tags]:
...
+            if timestamp is None:
+                now = arrow.now()
+                # honor the timestamp format, but default to date-only
+                if changelog and changelog[-1].extended_timestamp:
+                    timestamp = now.datetime
+                else:
+                    timestamp = now.date()
+            if author is None:
+                try:
+                    author = subprocess.check_output("rpmdev-packager").decode().strip()
```
FrNecas, line 176: This confused me a bit at first (setting email to None), maybe a comment mentioning that author from rpmdev-packager already contains the formatted mail could be useful.

specfile/changelog.py
nforro: No, the `get()` method actually searches for a match within the whole string (admittedly it's not made obvious in the official documentation, apart from several code examples), so it works as expected.

specfile/specfile.py
nforro: I'll do that. In fact, I'd prefer not separating e-mail and author's name and treat it as a single entity (RPM does so as well), but it was a requirement from packit.

FrNecas: Good idea, I think that would be fine, the ChangelogEntry class will be simple and packit will have a high-level API for setting author and email separately. I was going to propose one more alternative approach (having an `Author` class that could be constructed either in the format of rpmdev-packager or from name and email) but that seems to be too complicated, especially from the caller's point of view so I'd go with what you proposed

FrNecas (approved):

PR #5, opened 2022-02-03
Some fixes and integration test for Specfile.add_changelog_entry()

nforro:


specfile/tags.py
```diff
@@ -333,6 +333,20 @@ def __contains__(self, item: object) -> bool:
...
+    def __getitem__(self, i: int) -> Comment:
+        pass
+
+    @overload
+    def __getitem__(self, i: slice) -> "Comments":
+        pass
+
+    def __getitem__(self, i):
+        if isinstance(i, slice):
+            return Comments(self.data[i], self._preceding_lines)
```
FrNecas, line 346: I'm wondering about the `preceding_lines` here. Does it make sense if the comments before the tag look something like

```
preceding 1
preceding 2
# comment 1
# comment 2
# comment 3
```

and the slice is something like `[1:]`? This would still pass the 2 preceding lines to the constructor which doesn't make much sense since they are not preceding the comments directly. On the other hand, I wouldn't expect anyone to slice the comments and then operate with the preceding lines (e.g. assembling or something similar) so it's probably not an issue.

FrNecas: The same goes for Tags a bit below (with `remainder)`

tests/integration/test_specfile.py
```diff
@@ -0,0 +1,111 @@
+# Copyright Contributors to the Packit project.
+# SPDX-License-Identifier: MIT
+
+import datetime
+import shutil
+import subprocess
+from pathlib import Path
+
+import pytest
+from flexmock import flexmock
```
FrNecas, line 10: As far as I can see, `flexmock` is not in `setup.cfg`, you should probably add it there

FrNecas (commented):
Just one question, otherwise looks good to me.

specfile/tags.py
nforro: Good point. I agree that it doesn't make much sense. Perhaps slicing should simply eliminate preceding lines/remainder?

tests/integration/test_specfile.py
nforro: Will do.

specfile/tags.py
FrNecas: Good point. Depends on how you look at slicing and what it may be useful for in this case. Lets keep it then, at least for now, we can always change it later if we find it to be problematic

FrNecas (approved):

PR #7, opened 2022-02-09
Release 0.1.0

nforro:


.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,29 @@
...
+        # https://pypa-build.readthedocs.io
+        run: |
+          python -m pip install build
+          python -m build --sdist --wheel
+
+      - name: Publish 📦 to PyPI
+        # https://github.com/pypa/gh-action-pypi-publish
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          password: ${{ secrets.PYPI_API_TOKEN }}
```
FrNecas, line 28: I assume the secret is set on org level, but we should verify that just to make sure.

FrNecas (approved):

.github/workflows/pypi-publish.yml
nforro: As far as I can tell, it's not, I actually set my own token on repo level.

nforro: It's a temporary one anyway, as the project on PyPI doesn't exist yet.

FrNecas: Oh, I remember now, the token is not tied to an account but rather a project IIRC. All good then :)

PR #8, opened 2022-02-15
Enable Zuul CI for the project

FrNecas:
- added Containerfile for testing
- added Makefile with testing-related targets
- playbooks for ZuulCI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,18 @@
+TEST_IMAGE = specfile-tests
+CONTAINER_ENGINE ?= $(shell command -v podman 2> /dev/null || echo docker)
+
+TEST_TARGET ?= ./tests/unit ./tests/integration
+
+.PHONY: check install build-test-image check-in-container
+
+check:
+	PYTHONPATH=$(CURDIR) PYTHONDONTWRITEBYTECODE=1 python3 -m pytest --verbose --showlocals $(TEST_TARGET) --full-trace
+
+install:
+	pip3 install --user .
+
+build-test-image:
+	$(CONTAINER_ENGINE) build --rm --tag $(TEST_IMAGE) -f Containerfile.tests
+
+check-in-container:
+	$(CONTAINER_ENGINE) run --rm -ti -v $(CURDIR):/src:Z -w /src --env TEST_TARGET $(TEST_IMAGE) make check
--- a/files/tasks/rpm-test-deps.yaml
+++ b/files/tasks/rpm-test-deps.yaml
@@ -1,0 +1,7 @@
+---
+- name: Install test RPM dependencies
+  dnf:
+    name:
+      - python3-flexmock
+      - python3-pytest
+  become: true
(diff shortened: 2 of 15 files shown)


specfile/exceptions.py
```diff
@@ -23,3 +23,10 @@ def __str__(self) -> str:
             if line.startswith("error:"):
                 return line.split("error: ")[1]
         return "\n" + b"".join(self.stderr).decode()
+
+
+class RPMMacroRemovalException(RPMException):
+    """Exception related to failed removal of RPM macros."""
+
+    def __init__(self) -> None:
+        super().__init__([])
```
nforro, line 32: Wouldn't it be better to either inherit from more generic exception or to modify `RPMException` so that the `stderr` attribute is optional? This would traceback with a newline, which could be a bit confusing, and I don't like it in general.
How about just `MacroRemovalException` inheriting from `SpecfileException`?

specfile/rpm.py
```diff
@@ -180,15 +182,26 @@ def remove(cls, macro: str) -> None:
...
             rpm.delMacro(macro)
             try:
                 if cls.expand(f"%{macro}") == f"%{macro}":
                     break
             except RPMException:
                 # the macro can't be expanded, but it still exists
+                retry += 1
                 continue
+        else:
+            raise RPMMacroRemovalException
```
nforro, line 204: Some message would be nice, maybe something like `Max attempts exceeded`.

tests/unit/test_rpm.py
```diff
@@ -82,15 +84,23 @@ def test_macros_remove():
...
 
+@pytest.mark.xfail(
+    rpm.__version__.split(".")[0] == "4" and int(rpm.__version__.split(".")[1]) < 17,
+    reason="Expansion and removal of built-in parametrized macros is broken before rpm 4.17",
+    raises=RPMMacroRemovalException,
+)
+def test_macros_remove_builtin_parametrized():
+    rpm.reloadConfig()
+    Macros.remove("echo")
+    assert Macros.expand("%echo abcd") == "%echo abcd"
```
nforro, line 95: I'd parametrize this to try to remove also a macro defined with `rpm.addMacro()`. Thinking about it, it would be great to test also some rpmrc macros, both simple and parametrized, however that would require some non-trivial setup, so it's more suitable for an integration test.

Makefile
```diff
@@ -0,0 +1,18 @@
+TEST_IMAGE=specfile-tests
```
nforro, line 1: Just a nitpick, this is not a shell script, so whitespace around `=` is allowed (and it would make this consistent with the two assignments below).

nforro, line 1: Just a nitpick, this is not a shell script, so whitespace around `=` is allowed (and it would make this consistent with the two assignments below).

files/tasks/rpm-test-deps.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-- name: Install test rpm dependencies
+- name: Install test RPM dependencies
```
nforro, line 2: It would be nice to move this change to the previous commit.

nforro (approved):

nforro:
@FrNecas sorry, I forgot you wanted to test the `mergeit` label

FrNecas:
Np, lets try next time but I think it should work fine since zuul seems to be working :)

PR #9, opened 2022-02-16
Drop runtime dependency on setuptools

hroncok:
The setuptools package is a huge overkill to get self-version.

Use the standard library instead on Python 3.8+,
or a smaller package on older Pythons.

Declare support for new Pythons (extra commit).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,6 +20,8 @@ classifiers =
     Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
     Topic :: Software Development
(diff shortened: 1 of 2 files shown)


hroncok:
I am not sure I understand the mypy failure. Should I noqa it?

nforro:
> I am not sure I understand the mypy failure. Should I noqa it?

It seems to be a known issue: https://github.com/python/mypy/issues/1153

hroncok:
Added `# type: ignore`

hroncok:
But the pre-commit thing now made it uglier :(

hroncok:
I've tried really hard to make it both readable and acceptable by the robot overlords.

nforro (approved):
Thanks!

TomasTomecek:
Very nice, I didn't know there is a more lightweight way to do this.

PR #10, opened 2022-02-16
Some fixes and updates

nforro:
* Make `Specfile` importable directly from the package
* Fix parsing `Source` and `Patch` tags
* Update Fedora spec file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,4 @@
+# 0.1.1
+
+- Fixed parsing _Source_ and _Patch_ tags
+- Made `Specfile` importable directly from the package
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# 0.1.1
+
+## Fixes
```
FrNecas, line 3: Hm, do we want to structure it like this? Looking at other packit repos, the changes are always directly under the tag header so I'd maybe follow this for consistency.

FrNecas (approved):
Just a consistency nitpick

CHANGELOG.md
nforro: Oh, right, I was looking at some older changelog entries and didn't realize the style has changed.

PR #23, opened 2022-03-16
Allow to fully parse spec files with missing local sources

### LINKED ISSUE #19: Enable spec file parsing without remote sources/patches present
To be able to download remote sources/patches, it is necessary to be able to parse the spec file without them present.
This is a prerequisite for #13.

nforro:
RPM is unable to fully parse a spec file if some sources are missing. Only so called non-build parse is possible in that case, but the result is not the same, it's like the sources weren't even defined.

Overcome that by creating a temporary dummy source for every source that is missing locally. Mimic file signatures of compressed sources to make RPM behave exactly the same when processing a dummy source.

Fixes #19.


specfile/rpm.py
```diff
@@ -236,6 +237,51 @@ def reinit(cls, level_threshold: MacroLevel = MacroLevel.RPMRC) -> None:
...
+            (".bz2", b"BZh"),
+            (".zip", b"PK00"),
+            (".xz", b"\xfd\x37\x7a\x58\x5a\x00"),
+            (".zst", b"\x28\xb5\x2f"),
+            (".lz", b"LZIP"),
+            (".lrz", b"LRZI"),
+            (".gz", b"\x1f\x8b"),
+            (".7z", b"7z\xbc\xaf\x27\x1c"),
+        ]
+        MAGIC_LENGTH = 13
```
FrNecas, line 265: What does magic length mean? Do I understand it correctly that it's the number of characters that rpm takes a look at while parsing? If so, a comment would be nice

FrNecas (approved):
Nice workaround!

FrNecas:
As you mentioned earlier, zuul doesn't seem to like having target branch as something else than main/master. The `mergeit` label won't work (it requires the zuul check job to pass) so you will have to merge manually via Github

specfile/rpm.py
nforro: Yes, RPM reads the first 13 bytes of a file to do signature matching. I thought it was self-explanatory, but I'll add a comment.

nforro:
> As you mentioned earlier, zuul doesn't seem to like having target branch as something else than main/master. The `mergeit` label won't work (it requires the zuul check job to pass) so you will have to merge manually via Github

It should work after I change the base branch (after PR #20 is merged).

PR #25, opened 2022-03-22
Configure Packit for the python-specfile package

lachmanfrantisek:
There are multiple jobs configured:

* `propose_downstream` for release => will submit pull-requests for a new GitHub release.
* `copr_build` for PR => will trigger the Copr build for pull-requests.
* `copr_build` for `main` branch commits and releases => Will built in the Packit's Copr projects
  where other projects like `python-ogr` and `packit` are built.
* `koji_build` for new dist-git commits for all Fedora branches and `epel-8`.
* `bodhi_update` for successful builds for all stable Fedora branches and `epel-8`.

Signed-off-by: Frantisek Lachman <flachman@redhat.com>

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,82 @@
+---
+specfile_path: fedora/python-specfile.spec
+
+# add or remove files that should be synced
+files_to_sync:
+  - fedora/python-specfile.spec
+  - .packit.yaml
+
+# name in upstream package repository or registry (e.g. in PyPI)
+upstream_package_name: specfile
+# downstream (Fedora) RPM package name
+downstream_package_name: python-specfile
+
+copy_upstream_release_description: true
+
+actions:
+  # we need this b/c `git archive` doesn't put all the metadata in the tarball:
+  #   LookupError: setuptools-scm was unable to detect version for '/builddir/build/BUILD/ogr-0.11.1'.
+  #   Make sure you're either building from a fully intact git repository or PyPI tarballs.
+  create-archive:
+    - python3 setup.py sdist --dist-dir ./fedora/
+    - bash -c "ls -1t ./fedora/*.tar.gz | head -n 1"
+  get-current-version: python3 setup.py --version
+
+srpm_build_deps:
+  - python3-pip # "python3 setup.py --version" needs it
+  - python3-setuptools_scm
+
+jobs:
+  - job: propose_downstream
+    trigger: release
+    metadata:
+      dist_git_branches:
+        - fedora-all
+        - epel-8
+
+  - job: copr_build
+    trigger: pull_request
+    metadata:
+      targets:
+        - fedora-all
+        - epel-8
+  - job: tests
+    trigger: pull_request
+    metadata:
+      targets:
+        - fedora-all
+        - epel-8
+
+  - job: copr_build
+    trigger: commit
+    metadata:
+      branch: main
+      targets:
+        - fedora-all
+        - epel-8
+      project: packit-dev
+      list_on_homepage: True
+      preserve_project: True
+  - job: copr_build
+    trigger: release
+    metadata:
+      targets:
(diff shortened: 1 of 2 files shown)


nforro:
Thanks! I suppose I'll have to make some spec file changes for EPEL 8, but let's see :)

nforro:
> Thanks! I suppose I'll have to make some spec file changes for EPEL 8, but let's see :)

Talking about this, is it possible to have a different `specfile_path` for EPEL 8?

lachmanfrantisek:
> Talking about this, is it possible to have a different `specfile_path` for EPEL 8?

Sure, if we need to, we can define another job for that. Let me show that in another commit.

lachmanfrantisek:
Ok, looks like `Copr` does not like the multi-spec scenario..;(

```
Copr build error: too many specfiles: /var/lib/copr-rpmbuild/workspace/workdir-bswcfp_2/srcdir/python-specfile-epel-8.spec, /var/lib/copr-rpmbuild/workspace/workdir-bswcfp_2/srcdir/python-specfile.spec
```

.packit.yaml
```diff
@@ -32,19 +32,40 @@ jobs:
     metadata:
       dist_git_branches:
         - fedora-all
+  - job: propose_downstream
+    trigger: release
+    files_to_sync:
+      # There is a different spec-file for epel-8:
+      - fedora/python-specfile-epel-8.spec
```
nforro, line 39: I was thinking about something like `epel/python-specfile.spec`, but that's only a detail.

nforro: But maybe it would help with the Copr issue?

lachmanfrantisek: OK, I'll try that.

TomasTomecek:
> Ok, looks like `Copr` does not like the multi-spec scenario..;(
> 
> ```
> Copr build error: too many specfiles: /var/lib/copr-rpmbuild/workspace/workdir-bswcfp_2/srcdir/python-specfile-epel-8.spec, /var/lib/copr-rpmbuild/workspace/workdir-bswcfp_2/srcdir/python-specfile.spec
> ```

ran into this as well with leapp-repo; had to delete or move the other spec file

lachmanfrantisek:
@nforro it looks like we might hit some issues when doing two chroot updates in a short period of time (because of the two jobs defined). Would it be possible to use one spec-file for now?

nforro:
> @nforro it looks like we might hit some issues when doing two chroot updates in a short period of time (because of the two jobs defined). Would it be possible to use one spec-file for now?

Possible, yes, but it means "downgrading" from pyproject macros, which makes it a bit difficult to comply with Fedora Packaging Guidelines. But we need to support EPEL 8, so I guess there is no way around it.

lachmanfrantisek:
Sorry for the mess and all the commits. I am trying to find an issue with this setup.

I will let you know once it's ready...

lachmanfrantisek:
@nforro ok, now it should work well -- it just uses a different Copr project for Epel and Fedora if it is not an issue. I've squashed the commits so you can merge this as is and fix the Epel specfile yourself if it is ok for you.

Or, do you want to change/tweak anything else?

nforro (approved):
This looks great, thank you!

nforro:
> @nforro ok, now it should work well -- it just uses a different Copr project for Epel and Fedora if it is not an issue. I've squashed the commits so you can merge this as is and fix the Epel specfile yourself if it is ok for you.

Great, I'll update the EPEL spec file. Should I do it in this PR? Is that even possible?

> Or, do you want to change/tweak anything else?

I can't think of anything.

nforro:
> I think you should be able to push to my branch, but you can also merge this PR and send a follow-up with the fix.

Ok, let me merge this and open a new PR, I'll update the Fedora spec file as well.

PR #29, opened 2022-03-23
Add Sections.get() method

FrNecas:
This is an alternative syntax for accessing a section based on its name.
For cases where a section is parametrized, this may be more convenient
than using getattr(sections, 'files devel').


README.md
```diff
@@ -76,6 +76,7 @@ with specfile.sections() as sections:
     sections[1], sections[2] = sections[2], sections[1]
     # accessing a section with arguments
     print(getattr(sections, 'package devel'))
+    print(sections.get('package devel'))
```
nforro, line 79: I'd move this above `getattr` or perhaps leave out `getattr` completely.

nforro (approved):

PR #30, opened 2022-03-23
Make Sections copyable

FrNecas:
Previously, copying Sections would result in infinite recursion due to
accessing the data attribute through __getattr__. Avoid this recursion
by handling non-section attribute names in __getattr__ separately.


nforro:
Good catch, however, if I'm not mistaken the same issue is in every other class except for `Changelog`. Could you fix it everywhere please?

nforro (approved):

nforro:
I've opened #31.

FrNecas:
Good point that this is also the case in more places. I will push changes to the other classes to the PR a bit later. I am just wondering what the correct way to solve this is. `return super().__getattribute__(name)` this seems to work fine... `self.__getattribute__(name)` is probably the equivalent so it may be better than calling super?

FrNecas:
So Tags should be fixed, the last part that can't be deep-copied is the `Specfile` class which fails due to `rpm.spec`:

```
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.9/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/usr/lib64/python3.9/copy.py", line 270, in _reconstruct
    state = deepcopy(state, memo)
  File "/usr/lib64/python3.9/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/usr/lib64/python3.9/copy.py", line 230, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/usr/lib64/python3.9/copy.py", line 161, in deepcopy
    rv = reductor(4)
TypeError: cannot pickle 'rpm.spec' object
```

I have to take a look if there is a way to overcome this.

nforro:
> So Tags should be fixed, the last part that can't be deep-copied is the `Specfile` class which fails due to `rpm.spec`:

I think this is fine, we don't need to support copying `Specfile`.

specfile/tags.py
```diff
@@ -463,6 +463,17 @@ def __init__(
...
 
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Tag):
+            return NotImplemented
+        return (
+            self.name == other.name
+            and self.value == other.value
+            and self._expanded_value == other._expanded_value
+            and self._separator == other._separator
+            and self.comments == other.comments
```
nforro, line 474: Do we consider `Tag` to be mutable? Wouldn't it make sense to implement also `__hash__()`, like in `Macro`?
See: [https://docs.python.org/3.11/reference/datamodel.html#object.__hash__](https://docs.python.org/3.11/reference/datamodel.html#object.__hash__).

specfile/tags.py
```diff
@@ -539,6 +550,8 @@ def __getitem__(self, i):
             return self.data[i]
 
     def __getattr__(self, name: str) -> Tag:
+        if name.capitalize().rstrip("0123456789") not in TAG_NAMES:
+            return super().__getattribute__(name)
```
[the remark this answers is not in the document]
FrNecas: Yes, good catch, I messed up when rebasing :)

specfile/tags.py
FrNecas: I think we should consider them mutable. This also corresponds to how the `tags()` context manager in `Specfile` operates -- you can modify a value of a tag and then the change is mapped back to the low-level content. Whereas with macros, it doesn't make much sense to change anything.

nforro (approved):

PR #32, opened 2022-03-23
Update spec files

nforro:
This is a follow-up to #25.

The EPEL spec file should be correct now, however the build on EPEL fails because of multiple issues, namely:
* missing `tmp_path` fixture (already solved in [packit](https://github.com/packit/packit/blob/070792da802b32d29ce527838fe1c0c1137f5c79/tests/conftest.py#L28-L39))
* infinite loop in `Macros.remove()` (fixed in https://github.com/packit/specfile/commit/e3d3022b8781b104bc6b18aa770f43fe1e1d8476)
* older version of `arrow` inability to parse changelog entry headers (needs investigation)

### DIFF
--- a/epel/python-specfile.spec
+++ b/epel/python-specfile.spec
@@ -17,6 +17,9 @@ Source0:        %{pypi_source specfile}
 BuildArch:      noarch
 
 BuildRequires:  python%{python3_pkgversion}-devel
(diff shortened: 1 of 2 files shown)


jpopelka (approved):

nforro:
I'm going to merge this and open a follow-up PR with fixes for EPEL 8.

PR #37, opened 2022-03-30
Implement mid-level manipulation of %prep section

### LINKED ISSUE #17: Implement mid-level manipulation of %prep section
Necessary functionality:
* Detecting "format" (`%autosetup`, `%setup` + `%autopatch`, `%setup` + `%patch`, something else)
* Adding/removing `%patch` macros
* Parsing macro options

nforro:
Fixes #17.


nforro:
I'm working on tests and I want to write some usage examples, otherwise it should be pretty much done.

specfile/prep.py
```diff
@@ -0,0 +1,303 @@
...
+    def _get_items(self) -> List[PrepMacro]:
+        """
+        Gets all supported %prep macros.
+
+        Returns:
+            List of instances of subclasses of PrepMacro.
+        """
+        comment_regex = re.compile(r"^\s*#.*$")
+        macro_regex = re.compile(
+            r"(?P<c>%{!?\?\w+:)?.*?"
```
FrNecas, line 158: I'm a bit lost in this regex. The second group (the macro itself) and the last (options) are clear but what about the first group?

specfile/prep.py
```diff
@@ -0,0 +1,303 @@
...
+            if comment_regex.match(line):
+                continue
+            m = macro_regex.search(line)
+            if not m:
+                continue
+            name, options = m.group("m"), m.group("o")
+            macro: PrepMacro
+            if name == "%setup":
+                macro = SetupMacro(i, name)
+                macro._parse_options(options)
```
FrNecas, line 173: Can't we take these _parse_options and append call out of all the ifs and have it here once for all the branches? The only exception will be the `%patch` special-case.

FrNecas, line 173: Can't we take these _parse_options and append call out of all the ifs and have it here once for all the branches? The only exception will be the `%patch` special-case.

specfile/prep.py
```diff
@@ -0,0 +1,303 @@
...
+        if p is not None:
+            macro += f" -p{p}"
+        if b is not None:
+            macro += f" -b{b}"
+        if E:
+            macro += " -E"
+        macros = [m for m in self.macros if isinstance(m, PatchMacro)]
+        if macros:
+            lines = [
+                m._line
```
FrNecas, line 276: Not a big fan of accessing the "private" attribute here. Perhaps `line` should be public?

FrNecas, line 276: Not a big fan of accessing the "private" attribute here. Perhaps `line` should be public?

specfile/prep.py
```diff
@@ -0,0 +1,303 @@
...
+                for m in sorted(macros, key=lambda m: m.index)
+                if m.index < index
+            ]
+            if lines:
+                self._section.insert(lines[-1] + 1, macro)
+            else:
+                self._section.insert(macros[0]._line, macro)
```
FrNecas, line 283: Doesn't this go against the docstring? This looks to me like it is inserted at the beginning

FrNecas, line 283: Doesn't this go against the docstring? This looks to me like it is inserted at the beginning

FrNecas:
Eh, github for some reason duplicated my comment, sorry for that!

specfile/prep.py
nforro: That's the start of a conditionalized macro expansion, it needs to be there to be able to match the ending bracket to ensure it's not included in the option string.

Real world example: `%{?with_system_nss:%patch30 -p3 -b .nss_pkcs11_v3}`

I'll add a comment.

specfile/prep.py
nforro: Ok, let me try that.

specfile/prep.py
nforro: The docstring says:
> If there are existing %patch macros,  the new macro is added before, after or between them according to index

This is the *before* case.

nforro:
@FrNecas I marked this as ready for review, but later realized it will probably be necessary to implement also modification, or at least reindexing of `%patch` macros. Should do it in this PR or open a new one?

FrNecas:
Either is fine with me, up to you ;) But this PR is getting a little large so if you don't expect the modifications to require significant changes to the current implementation, it would make sense to review and merge this and then open a new PR

FrNecas (approved):
LGTM, just 2 small nits

README.md
```diff
@@ -113,6 +113,20 @@ with specfile.changelog() as changelog:
...
+
+# working with macros in %prep section, supports %setup, %patch, %autosetup and %autopatch
+with specfile.prep() as prep:
+    # name of the first macro
+    print(prep.macros[0].name)
+    # options of the last macro
+    print(prep.macros[-1].options)
+    # checking if %autosetup is being used
+    print('%autosetup' in prep.macros)
+    print(AutosetupMacro in prep.macros)
```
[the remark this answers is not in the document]
nforro: Good point, I'll fix that.

tests/integration/test_specfile.py
```diff
@@ -23,6 +24,31 @@ def test_parse(spec_multiple_sources):
...
+            assert isinstance(m, PatchMacro)
+            assert m.index == i
+            assert m.options.p == 1
+        prep.remove_patch_macro(0)
+        assert len([m for m in prep.macros if isinstance(m, PatchMacro)]) == 2
+        prep.add_patch_macro(0, p=2, b=".test")
+        assert len(prep.macros) == 4
+    with spec.sections() as sections:
+        assert sections.prep[1] == "%patch0 -p2 -b .test"
+    spec = Specfile(spec_autosetup)
```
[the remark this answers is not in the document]
nforro: Ok, makes sense.

PR #56, opened 2022-04-27
Add support for %autochangelog

### LINKED ISSUE #53: Support %autochangelog
It is necessary to detect `%autochangelog`, `add_changelog_entry()` should be noop in that case.
Also, currently, accessing changelog using the `changelog()` context manager empties the entire `%changelog` section when `%autochangelog` is being used.

nforro:
Fixes #53.

### DIFF
--- a/specfile/specfile.py
+++ b/specfile/specfile.py
@@ -224,6 +224,17 @@ class Specfile:
                 for section, patchlist in patchlists:
                     section.data = patchlist.get_raw_section_data()
 
+    @property
+    def has_autochangelog(self) -> bool:
+        """Whether the spec file uses %autochangelog."""
+        with self.sections() as sections:
+            try:
+                section = sections.changelog
+            except AttributeError:
+                return False
+            changelog = [ln.strip() for ln in section if ln.strip()]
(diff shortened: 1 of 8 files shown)


specfile/specfile.py
```diff
@@ -265,4 +279,8 @@ def add_changelog_entry(
                     raise SpecfileException("Failed to auto-detect author") from e
             elif email is not None:
                 author += f" <{email}>"
-            changelog.append(ChangelogEntry.assemble(timestamp, author, entry, evr))
+            changelog.append(
+                ChangelogEntry.assemble(
+                    timestamp, author, entry, evr, append_newline=bool(changelog)
```
FrNecas, line 284: Shouldn't part of this diff be in the second commit? In `Add support for %autochangelog`, there is no `append_newline` yet.

FrNecas (approved):
One nitpick related to belonging to commits, otherwise looks good

specfile/specfile.py
nforro: Yes, it does, I'll fix it.

PR #69, opened 2022-07-20
Various fixes and improvements

nforro:
A bunch of fixes and improvements mostly based on feedback from https://github.com/packit/packit/pull/1588


tests/unit/test_sources.py
```diff
@@ -109,11 +109,39 @@ def test_sources_get_initial_tag_setup(tags, number, index):
...
-            ["Source2", "Source3", "Source4", "Source5"],
+            ["Source0", "Source0", "Source", "Source"],
+            ["Source0", "Source1", "Source", "Source"],
+        ),
+        (
+            ["Source100", "Source100", "Source", "Source1028", "Source1011"],
+            ["Source100", "Source101", "Source", "Source1028", "Source1011"],
+        ),
+        (
+            ["Source2", "Source2", "Source3", "Source", "Source999"],
```
FrNecas, line 124: What if this was `Source5` rather than `Source999`? The current implementation would change it to `Source` rather than `Source6`. Is this expected or a bug?

specfile/sources.py
```diff
@@ -100,14 +101,26 @@ def _extract_number(self) -> Optional[str]:
...
+        has no leading zeros and the actual number of digits if there are
+        any leading zeros.
+        """
+        if self._number is not None:
             return 0
-        return len(self._extract_number() or "")
+        number = self._extract_number()
+        if not number:
+            return 0
+        zeros = list(itertools.takewhile(lambda c: c == "0", number))
```
FrNecas, line 122: Wouldn't `number.startswith("0")` suffice? We don't need the actual number of zeros.

specfile/tags.py
```diff
@@ -340,14 +341,29 @@ def __getitem__(self, i):
 
     def __delitem__(self, i: Union[SupportsIndex, slice]) -> None:
```
FrNecas, line 347: Could you add at least a simple test case for the compression?

FrNecas (commented):
Looks good, a few suggestions

specfile/sources.py
nforro: You are right, I'll fix that.

specfile/tags.py
nforro: It won't be so simple, but sure, I will :slightly_smiling_face:

FrNecas (approved):
Thanks for all the improvements, looks good now!

PR #77, opened 2022-08-03
Fix detecting extended timestamp format in %changelog

### LINKED ISSUE #76: Fix parsing timestamps in changelog
Changelog entry `* Fri Jun 3 2022 Eric Gustavsson <egustavs@redhat.com> 0.26` is considered to have extended timestamp format, which is obviously wrong. The reason is that `arrow` fails to parse the timestamp using `ddd MMM DD YYYY` pattern.

nforro:
Fixes #76.

### DIFF
--- a/specfile/changelog.py
+++ b/specfile/changelog.py
@@ -55,7 +55,7 @@ class ChangelogEntry:
     def extended_timestamp(self) -> bool:
         """Whether the timestamp present in the entry header is extended (date and time)."""
         try:
-            arrow.get(self.header, "ddd MMM DD YYYY")
+            arrow.get(self.header, ["ddd MMM D YYYY", "ddd MMM DD YYYY"])
         except arrow.parser.ParserError:
             return True
         else:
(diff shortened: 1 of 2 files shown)


specfile/changelog.py
```diff
@@ -55,11 +55,13 @@ def __repr__(self) -> str:
     def extended_timestamp(self) -> bool:
         """Whether the timestamp present in the entry header is extended (date and time)."""
         try:
-            arrow.get(self.header, "ddd MMM DD YYYY")
+            arrow.get(self.header, "ddd MMM D YYYY")
```
FrNecas, line 58:
```suggestion
            arrow.get(self.header, ["ddd MMM D YYYY", "ddd MMM DD YYYY"])
```

Arrow allows passing a list of formats to try, the functionality should be equivalent but the code is simpler IMO.

nforro: Thanks, that's definitely better.

FrNecas (approved):

PR #91, opened 2022-08-30
Build specfile for EPEL 9

nforro:


plans/git_reference.py
```diff
@@ -0,0 +1,14 @@
...
+
+import subprocess
+from pathlib import Path
+
+import fmf
+
+tree_root = Path.cwd().absolute()
+node = fmf.Tree(tree_root).find("/plans")
+with node as data:
+    data["discover"]["repository"] = "https://github.com/packit/specfile.git"
```
FrNecas, line 11: Use `url` instead of `repository`. The latter is only supported for backwards-compatibility. I am not sure if or when it will be removed but the [tmt docs](https://tmt.readthedocs.io/en/stable/spec/plans.html#fmf) only mention `url` so I'd go with that to be safe for the future.

plans/smoke.fmf
```diff
@@ -0,0 +1,4 @@
+summary:
+  Basic smoke test.
```
FrNecas, line 2: Nitpick: I'd suggest not using full stops at the end of summaries. I personally view it similarly to commit messages and the first lines but it's more of a personal taste :)

plans/full.fmf
```diff
@@ -0,0 +1,10 @@
+summary:
+  Unit & integration tests.
+discover+:
+  filter: tier:1
+prepare:
```
FrNecas, line 5: I'd suggest using tmt's `install` method to avoid calling `dnf` directly. It handles other things such as running sudo if necessary and should also work on silverblue:

```yaml
prepare:
  - name: pip
    how: install
    package: python3-pip
  - name: flexmock
    how: shell
    # flexmock is not in EPEL 9, install it from PyPI
    script: pip3 install flexmock
```

FrNecas (commented):
Looks good, couple of possible improvements for the FMF tests

plans/git_reference.py
nforro: My bad, I saw it in some example and assumed that it's `url` that is deprecated (or less specific).

plans/full.fmf
nforro: Thanks, this is much better.

FrNecas (approved):

PR #98, opened 2022-09-08
Push to pypi when release is released

majamassarini:


majamassarini:
Already in specfile/#96

PR #100, opened 2022-09-12
Add support for filenames in source URL fragments

nforro:
Fixes [this](https://sentry.io/organizations/red-hat-0p/issues/3573060909/?alert_rule_id=855946&alert_timestamp=1662661800059&alert_type=email&environment=prod&project=1767823&referrer=alert_email) Sentry issue. Even though the issue is actually caused by a typo in [Source0 URL](https://src.fedoraproject.org/rpms/greenboot/blob/d7abb8237f1e0132af47c44717c92f5afd488015/f/greenboot.spec#_14), the typo goes unnoticed by standard tooling because the correct filename is specified in URL fragment. 

---

RELEASE NOTES BEGIN
Added support for filenames specified in source URL fragments, for example: `https://example.com/foo/1.0/download.cgi#/%{name}-%{version}.tar.gz`
RELEASE NOTES END

### DIFF
--- a/specfile/rpm.py
+++ b/specfile/rpm.py
@@ -8,7 +8,6 @@ import os
 import re
 import sys
 import tempfile
-import urllib.parse
 from enum import IntEnum
 from pathlib import Path
 from typing import Iterator, List, Optional, Tuple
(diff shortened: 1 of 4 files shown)


lachmanfrantisek (approved):
Wow. I haven't known this syntax exists...

PR #104, opened 2022-10-05
Add support for filtering changelog entries by EVR

nforro:
Related to https://github.com/packit/packit-service/issues/1518.

RELEASE NOTES BEGIN
It is now possible to filter changelog entries by specifying lower bound EVR, upper bound EVR or both.
RELEASE NOTES END

### DIFF
--- a/specfile/changelog.py
+++ b/specfile/changelog.py
@@ -6,6 +6,8 @@ import datetime
 import re
 from typing import List, Optional, Union, overload
 
+import rpm
+
 from specfile.sections import Section
 from specfile.types import SupportsIndex
 
@@ -65,6 +67,27 @@ class ChangelogEntry:
         content = repr(self.content)
         following_lines = repr(self._following_lines)
         return f"ChangelogEntry('{self.header}', {content}, {following_lines})"
+
+    @property
+    def evr(self) -> Optional[str]:
+        """EVR (Epoch, Version, Release) of the entry."""
+        m = re.match(
+            r"""
+            ^.*
+            \s+                       # preceding whitespace
+            ((?P<sb>\[)|(?P<rb>\())?  # optional opening bracket
+            (?P<evr>(\d+:)?\S+-\S+?)  # EVR
+            (?(sb)\]|(?(rb)\)))       # matching closing bracket
+            :?                        # optional colon
+            \s*                       # optional following whitespace
+            $
+            """,
+            self.header,
+            re.VERBOSE,
+        )
+        if not m:
+            return None
+        return m.group("evr")
 
     @property
     def extended_timestamp(self) -> bool:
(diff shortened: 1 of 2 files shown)


specfile/changelog.py
```diff
@@ -223,6 +247,22 @@ def delete(index):
         else:
             delete(i)
 
+    def get_entries_since(self, evr: str) -> List[ChangelogEntry]:
```
lachmanfrantisek, line 250: What about implementing both limits (`from`/`to`, `before`/`after`, `since`/`to`...) and having `to`/`before` as optional with the current EVR as a default.

lachmanfrantisek (approved):

specfile/changelog.py
nforro: It took me a while to get what you mean :slightly_smiling_face: To be sure, are you talking about having two arguments (e.g. since, until) with defaults instead of one (evr)?

FrNecas: I think that would make sense, we could make it work a bit like list slicing. E.g. you would be able to specify: give me all changelog entries between 1.2.3 and 2.3.1. Both of these arguments could be optional meaning that one could use `since=None, until='1.2.3'` which would mean all releases until `1.2.3`.

nforro: I agree that it would be nice, I'm just not 100% sure this is what Franta had in mind.

lachmanfrantisek: Yes, exactly. That was what I had in mind.

>  Both of these arguments could be optional meaning that one could use since=None, until='1.2.3' which would mean all releases until 1.2.3.

+1, we can use beginning as a default for `since` and the current version for `until`.

specfile/changelog.py
```diff
@@ -223,6 +246,54 @@ def delete(index):
...
+        self, since: Optional[str] = None, until: Optional[str] = None
+    ) -> "Changelog":
+        """
+        Filters changelog entries with EVR between since and until.
+
+        Args:
+            since: Optional lower bound. If specified, entries with EVR higher
+              than this will be included.
+            until: Optional upper bound. If specified, entries with EVR lower
+              than this will be included.
```
lachmanfrantisek, line 259: +1

I am just not sure if it will be handier to have `higher` or `higher or equal`. What will be better for the Packit's usecase?

lachmanfrantisek (approved):
Exactly what I had in mind, thanks!

specfile/changelog.py
nforro: `higher` is more convenient, because we need a changelog diff since the last stable EVR. But that can be easily done with `higher or equal` and slicing. `higher or equal` is perhaps more intuitive?

nforro: Ok, let me change that.

lachmanfrantisek: Sorry for such nitpicking..;)

lachmanfrantisek (approved):

PR #114, opened 2022-10-20
Avoid using dataclasses with keyword-only fields

nforro:
Dataclasses with keyword-only fields are not in Python 3.9, avoid using them.

### DIFF
--- a/specfile/utils.py
+++ b/specfile/utils.py
@@ -1,21 +1,35 @@
 # Copyright Contributors to the Packit project.
 # SPDX-License-Identifier: MIT
 
+import collections
 import re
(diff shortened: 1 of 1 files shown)


TomasTomecek (approved):
I'd love to understand the change in the regexes :)

nforro:
> I'd love to understand the change in the regexes :)

Before an empty string was a match in case of EVR, and for example "0-" was also considered valid (as well as "name-0-" or "name-" or "-0" in case of NEVR).

mfocko:
> I'd love to understand the change in the regexes :)

It denies `--` in the version and requires at least one digit in between them

PR #113, opened 2022-10-18
Add utility dataclasses for working with (N)EVR

nforro:
This will also be useful in packit for working with NEVRs of Koji builds.

PR #116, opened 2022-10-20
Use class methods instead of static methods where convenient

nforro:
A follow-up to https://github.com/packit/specfile/pull/113#discussion_r999412293.


specfile/prep.py
```diff
@@ -350,25 +350,27 @@ def parse(section: Section) -> "Prep":
                     m.group("o"),
                 )
                 prefix, suffix = line[: m.start("m")], line[m.end("o") :]
-                cls = next(
+                class_ = next(
```
FrNecas, line 353: I believe that in p-s, we use `klass` instead of `class_` for things like this, I'd suggest to use that for consistency but that's just a nitpick.

FrNecas (approved):

specfile/prep.py
nforro: Ok :slightly_smiling_face:

PR #130, opened 2022-11-02
Add option to ignore missing includes

### LINKED ISSUE #121: Unable to open rpm-specs/coreutils-provides.inc: No such file or directory
I have very simple script:

```
#!/usr/bin/python3                                                                                             
from specfile import Specfile
import sys

# this is dumb, but do the work
filename = sys.argv[1]
specfile = Specfile(filename)

with specfile.sections() as sections:
    print(sections.changelog)

```
And when I run it as `./print-spec-changelog.py rpm-specs/coreutils.spec` I get:

```
  File "/usr/lib/python3.11/site-packages/specfile/rpm.py", line 319, in parse
    spec = rpm.spec(tmp.name, rpm.RPMSPEC_ANYARCH | rpm.RPMSPEC_FORCE)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: can't parse specfile


The above exception was the direct cause of the following exception:

    specfile = Specfile(filename)
               ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/specfile/specfile.py", line 48, in __init__
    self._spec = RPM.parse(str(self), self.sourcedir, self.macros)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/specfile/rpm.py", line 326, in parse
    raise RPMException(stderr=stderr) from e
specfile.exceptions.RPMException: Unable to open rpm-specs/coreutils-provides.inc: No such file or directory
```

The culprit is https://src.fedoraproject.org/rpms/coreutils/blob/rawhide/f/coreutils.spec#_13
Not sure who and why is trying to load the source file.

nforro:
Fixes #121.


specfile/rpm.py
```diff
@@ -284,9 +287,13 @@ def make_dummy_sources(sources: List[str], sourcedir: Path) -> Iterator[List[Pat
...
-    @staticmethod
+    @classmethod
     def parse(
-        content: str, sourcedir: Path, macros: Optional[List[Tuple[str, str]]] = None
+        cls,
+        content: str,
+        sourcedir: Path,
+        macros: Optional[List[Tuple[str, str]]] = None,
+        force: bool = False,
     ) -> rpm.spec:
```
FrNecas, line 297: Shouldn't this be `Tuple[rpm.spec, bool]`? I find it a bit weird that mypy is silent about this.

specfile/rpm.py
```diff
@@ -352,6 +352,7 @@ def get_included_sources(content):
             # do a non-build parse first, to get a list of sources
             spec = get_rpm_spec(content, rpm.RPMSPEC_ANYARCH | rpm.RPMSPEC_FORCE)
             sources = [s for s, _, _ in spec.sources]
+            del spec
```
FrNecas, line 355: I'd suggest adding a comment since it's quite unusual to see an explicit `del <object>` in Python code and a comment will make it easier to find the reason than a commit message

FrNecas (commented):
Looks good!

specfile/rpm.py
nforro: Thanks, yes, this was an oversight. I'm also wondering how come it passed mypy checks :thinking:

specfile/rpm.py
nforro: I thought about it, but since it's on multiple places and the explanation is quite long I wasn't sure how to deal with it. I'm thinking about maybe wrapping `rpm.spec` into a class that would handle this in one place?

FrNecas (approved):
Nice, thanks for adding the comment to the deletion! Do you have any idea why mypy didn't detect the previous typing error? I am just wondering if there are some mypy flags that we need to add.

nforro:
> Do you have any idea why mypy didn't detect the previous typing error? I am just wondering if there are some mypy flags that we need to add.

I didn't check, but I'll try to figure that out.

Issue #122, opened 2022-10-26
Empty tag: Provides


xsuchy:
specfile cannot parse https://src.fedoraproject.org/rpms/mingw-crt/blob/rawhide/f/mingw-crt.spec
It ends up with this TB:

```
The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/msuchy/projects/license-validate/./print-spec-changelog.py", line 7, in <module>
    specfile = Specfile(filename)
               ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/specfile/specfile.py", line 48, in __init__
    self._spec = RPM.parse(str(self), self.sourcedir, self.macros)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/specfile/rpm.py", line 326, in parse
    raise RPMException(stderr=stderr) from e
specfile.exceptions.RPMException: line 71: Empty tag: Provides:
```

nforro:
This is again caused by missing source files (`standard-dlls-mingw32` in this case).

nforro:
What we could theoretically do here is to generate a file containing some sort of a dummy value, but it would have to be a value that's syntactically correct in any place in a spec file, because shell expansions can appear anywhere. Let me think of something.

PR #134, opened 2022-11-09
Fake missing sources referenced from shell expansions

nforro:
Sources referenced from shell expansions are effectively equivalent to sources included using the `%include` directive, for instance `%include %{SOURCE1}` does the same thing as `%(cat %{SOURCE1})`.

Treat such sources that are not available the same way as missing included sources.

Related to #122.


nforro:
Actually, this is unnecessary, as missing sources referenced from shell expansions don't cause parsing to fail, RPM just emits a warning.

PR #137, opened 2022-11-14
Handle sources referenced from shell expansions in tag values

### LINKED ISSUE #122: Empty tag: Provides

nforro:
Sources referenced from shell expansions in tag values, for example `Summary: %(cat %{SOURCE1})`, are required even for non-build parse and can't be faked with empty files, because empty tag values are not allowed in a spec file.

If requested, fake such sources with a dummy value and mark the corresponding `Specfile` instance as tainted after doing so.

Also rename the `ignore_missing_includes` option to a more general `force_parse`.

Fixes #122.

RELEASE_NOTES_BEGIN
Renamed the `ignore_missing_includes` option to a more general `force_parse`. If specified, it allows to attempt to parse the spec file even if one or more sources required to be present at parsing time are not available. Such sources include sources referenced from shell expansions in tag values and sources included using the `%include` directive.
RELEASE_NOTES_END

---

Note:

There is another kind of sources that are required to be present even for non-build parse - macro files loaded with e.g. `%{load:%{SOURCE1}}`. But I don't think anyone would load a macro file in a spec without actually using any macros defined in it, so faking it doesn't really make sense, because parsing would fail on undefined macros anyway.


TomasTomecek:
> I'm wondering if this should be the default behavior (as it is currently implemented) or if there should be another option similar to `ignore_missing_includes` to explicitly enable this, or if we should reuse `ignore_missing_includes` for this as well and perhaps rename it.
> 
> There is another kind of sources that are required to be present even for non-build parse - macro files loaded with e.g. `%{load:%{SOURCE1}}`. But I don't think anyone would load a macro file in a spec without actually using any macros defined in it, so faking it doesn't really make sense, because parsing would fail on undefined macros anyway.

Makes sense to me to have a `force_parse=` argument that bypasses these checks and provides at least parts of a specfile that can be processed.

Hard to tell if it should be a default :)

nforro:
> Makes sense to me to have a `force_parse=` argument that bypasses these checks and provides at least parts of a specfile that can be processed.

Thanks, `force_parse` sounds good.

> Hard to tell if it should be a default :)

I don't think anything starting with `force` would be expected to be the default.

TomasTomecek:
> Since the `ignore_missing_includes` option already made its way to stable Fedoras and EPEL 9, should I keep it alongside the new `force_parse` option for backward compatibility? I'd rather avoid unnecessary clutter and we do say _This project is still a work in progress_ at the beginning of the README after all.

Makes sense to me to do a backwards incompatible change (and then don't update F36/F37).

I agree with you to have the 1.0 release with pleasant API :)

nforro:
> Makes sense to me to do a backwards incompatible change (and then don't update F36/F37).

Do you mean we should stop rebasing in stable Fedoras until 1.0 is released?

FrNecas (approved):

PR #147, opened 2022-12-01
Make sure expanded_release doesn't contain the dist suffix

nforro:
If the `Release` tag is defined by a macro, for example:

```spec
%global rel 28%{?dist}
Release:    %{rel}
```

then `Specfile.release` returns `%{rel}`, as expected, but, assuming `%dist` is defined as `.fc37`, `Specfile.expanded_release` returns `28.fc37` rather than `28`.
Fix that by explicitly defining `%dist` to be an empty string when expanding the release.

### DIFF
--- a/specfile/specfile.py
+++ b/specfile/specfile.py
@@ -530,7 +530,7 @@ class Specfile:
     @property
     def expanded_release(self) -> str:
         """Release string without the dist suffix with macros expanded."""
-        return self.expand(self.release)
+        return self.expand(self.release, extra_macros=[("dist", "")])
 
     def set_version_and_release(self, version: str, release: str = "1") -> None:
         """


specfile/specfile.py
```diff
@@ -530,7 +530,7 @@ def release(self, value: str) -> None:
     @property
     def expanded_release(self) -> str:
```
jpopelka, line 531: Shouldn't it be called `expanded_release_without_dist()` because I'd expect it returns `28.fc37` rather than `28` (unless I read the docstring first)

jpopelka (approved):

specfile/specfile.py
nforro: I decided to go with `release`/`expanded_release` and `raw_release`/`expanded_raw_release` pairs because the former will be used way more often and it's convenient to simply write for example `spec.release = 28`, without worrying about `%dist`.

nforro: On the other hand it's true that `spec.release_without_dist` would be completely self-explanatory.

nforro: I've opened https://github.com/packit/specfile/issues/148.

PR #150, opened 2022-12-03
Differentiate between sources and patches more

nforro:
While patches are technically sources, it can be a bit confusing to see for example a patch being represented by a `TagSource` instance. Improve on that.

This also makes it possible to differentiate between sources and patches when mixing both in one container.


jpopelka (approved):

nforro:
recheck

PR #153, opened 2022-12-08
Make context managers better

nforro:
This commit introduces a new `ContextManager` decorator that builds on top of `contextlib.contextmanager` and effectively makes context managers a singleton. Any nested context of the same kind will be the same as the outermost context of that particular kind.

This makes it possible, for example, to set properties that internally use context managers inside an existing context, or in general mix and nest context managers almost freely.

There are still instances of conflicting context managers, for example `macro_definitions()` and `sections()`, that, when combined, will overwrite each other's underlying data. All such instances will be documented.

Now it is also possible to access data directly, avoiding the `with` statement, by using `GeneratorContextManager.content` property (I'm open to suggestions how to name it better):

```python
for tag in spec.tags().content:
    print(f"{tag.name}: {tag.expanded_value}")
```

No modifications will be preserved though.

---

RELEASE NOTES BEGIN

Context managers (`Specfile.sections()`, `Specfile.tags()` etc.) can now be nested and combined together (with one exception - `Specfile.macro_definitions()`), and it is also possible to use tag properties (e.g. `Specfile.version`, `Specfile.license`) inside them. It is also possible to access the data directly, avoiding the `with` statement, by using the `content` property (e.g. `Specfile.tags().content`), but be aware that no modifications done to such data will be preserved. You must use `with` to make changes.

RELEASE NOTES END

### DIFF
--- a/specfile/context_management.py
+++ b/specfile/context_management.py
@@ -1,0 +1,141 @@
+# Copyright Contributors to the Packit project.
+# SPDX-License-Identifier: MIT
+
+import collections
+import contextlib
+import io
+import os
+import pickle
+import sys
+import tempfile
+import types
+from typing import Any, Callable, Dict, Generator, List, Optional, overload
+
+
+@contextlib.contextmanager
+def capture_stderr() -> Generator[List[bytes], None, None]:
+    """
+    Context manager for capturing output to stderr. A stderr output of anything run
+    in its context will be captured in the target variable of the with statement.
+
+    Yields:
+        List of captured lines.
+    """
+    fileno = sys.__stderr__.fileno()
+    with tempfile.TemporaryFile() as stderr, os.fdopen(os.dup(fileno)) as backup:
+        sys.stderr.flush()
+        os.dup2(stderr.fileno(), fileno)
+        data: List[bytes] = []
+        try:
+            yield data
+        finally:
+            sys.stderr.flush()
+            os.dup2(backup.fileno(), fileno)
+            stderr.flush()
+            stderr.seek(0, io.SEEK_SET)
+            data.extend(stderr.readlines())
+
+
+class GeneratorContextManager(contextlib._GeneratorContextManager):
+    """
+    Extended contextlib._GeneratorContextManager that provides get() method.
+    """
+
+    def __init__(self, function: Callable) -> None:
+        super().__init__(function, tuple(), {})
+
+    def __del__(self) -> None:
+        # make sure the generator is fully consumed, as it is possible
+        # that neither __enter__() nor content() have been called
+        collections.deque(self.gen, maxlen=0)
+
+    @property
+    def content(self) -> Any:
+        """
+        Fully consumes the underlying generator and returns the yielded value.
+
+        Returns:
+            Value that would normally be the target variable of an associated with statement.
+
+        Raises:
+            StopIteration if the underlying generator is already exhausted.
+        """
+        result = next(self.gen)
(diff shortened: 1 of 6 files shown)


nforro:
What do you think about the `get()` method? Other possibilities come to mind:

```python
# method
tags = spec.tags().get()
license = spec.tags().get().license

# property
tags = spec.tags().target
license = spec.tags().target.license

# GeneratorContextManager made callable
tags = spec.tags()()
license = spec.tags()().license

# GeneratorContextManager made iterable
tags = *spec.tags()
# no way to access properties directly
```

specfile/context_management.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def __init__(self, function: Callable) -> None:
+        super().__init__(function, tuple(), {})
+
+    def __del__(self) -> None:
+        # make sure the generator is fully consumed, as it is possible
+        # that neither __enter__() nor get() have been called
+        collections.deque(self.gen, maxlen=0)
+
+    def get(self) -> Any:
```
FrNecas, line 52: Out of the options that you mentioned, I probably like the explicit method the most, although I am not sure about the name. While `get()` is short, it is not very clear what it does (someone could expect it to be get method of dict). What about `.content()`?

specfile/context_management.py
```diff
@@ -0,0 +1,140 @@
...
+        Fully consumes the underlying generator and returns the yielded value.
+
+        Returns:
+            Value that would normally be the target variable of an associated with statement.
+
+        Raises:
+            StopIteration if the underlying generator is already exhausted.
+        """
+        result = next(self.gen)
+        next(self.gen, None)
```
FrNecas, line 63: What is the reason for the second `next` call? I am probably missing something

nforro: The generator yields a value (that's the first `next()`) and then does a clean-up (that's the second). It also raises `StopIteration`, that's why there is the second argument, `None`.

https://github.com/packit/specfile/blob/220170185359ffcd8583dcf7cd546f3c19f48afd/specfile/context_management.py#L131-L138

nforro: In theory in this case only the first `next()` would be sufficient, as the result is "read-only" anyway, but not exhausting the generator would leave it in suspended state, which would break [expectations](https://github.com/packit/specfile/blob/220170185359ffcd8583dcf7cd546f3c19f48afd/specfile/context_management.py#L110-L115), and also could cause troubles at destruction time (though that should be handled by the [destructor](https://github.com/packit/specfile/blob/220170185359ffcd8583dcf7cd546f3c19f48afd/specfile/context_management.py#L47-L50)).

specfile/context_management.py
FrNecas: Property also looks good (imho property/method are clearer than callable or iterable). In the end, it's a matter of taste :) Have you tried taking a look if some popular libraries do something similar so that we could get inspired there? That would probably be ideal.

nforro: > Have you tried taking a look if some popular libraries do something similar so that we could get inspired there?

Not really, do you know any that would employ this concept?

FrNecas: I first though of `requests` since they use some context managers but haven't found anything similar. Let's stick with `content` if that sounds good to you.

FrNecas (approved):

specfile/context_management.py
nforro: Let's go with `content` property, we can rename it later.

PR #157, opened 2022-12-14
Fix context managers being shared between Specfile instances

nforro:
This fixes use-cases such as this:

https://github.com/packit/packit/blob/d6488aac19417a0cf566ec284e86072d0b100cc2/packit/base_git.py#L409-L418


lbarcziova (approved):
thanks!

pcahyna:
I confirm that this change fixes the problem I saw in Packit (spec changes from upstream are not propagated to dist-git by the propose-downstream job).

PR #160, opened 2022-12-14
Use github-actions bot to run prepare-release steps

jpopelka:
Both, the [packit/prepare-release](https://github.com/peter-evans/create-pull-request#action-inputs) & [peter-evans/create-pull-request](https://github.com/packit/prepare-release#packitprepare-release) default to using (autogenerated) [secrets.GITHUB_TOKEN](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) if no token is provided.

That token should be enough because the steps are repo-scoped and [don't trigger any other workflow](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow) since the PR which triggers the 'Create GitHub release' is manually merged by a user.

Seems to work OK - jpopelka/packit#1


nforro (approved):
LGTM

Issue #163, opened 2022-12-15
Run packit tests on PRs to be sure nothing breaks


nforro:
- [ ] Add a new test plan that will *discover* and run packit tests

TomasTomecek:
We've done this with zuul for packit/ogr/p-s

I think now with TF it should be easier.

PR #164, opened 2022-12-16
Run Packit tests with new specfile changes

FrNecas:
Fixes: #163 

I hope that the RPM is installed in testing farm as an artifact, so this should work but let's see - I added a breaking change that should cause packit tests to fail

### DIFF
--- a/plans/git_reference.py
+++ b/plans/git_reference.py
@@ -5,10 +5,25 @@ from pathlib import Path
 
 import fmf
 
+# Set discover of specfile tests to a fixed commit
 tree_root = Path.cwd().absolute()
-node = fmf.Tree(tree_root).find("/plans")
-with node as data:
+tree = fmf.Tree(tree_root)
+main_node = tree.find("/plans")
+with main_node as data:
     data["discover"]["url"] = "https://github.com/packit/specfile.git"
     data["discover"]["ref"] = (
         subprocess.check_output(["git", "rev-parse", "HEAD"]).decode().strip()
+    )
+
+
+# Set discover of packit integration tests to the current main
+packit_node = tree.find("/plans/packit-integration")
+with packit_node as data:
+    data["discover"]["ref"] = (
+        subprocess.check_output(
+            ["git", "ls-remote", "https://github.com/packit/packit", "main"]
+        )
+        .decode()
+        .strip()
+        .split()[0]
     )
--- a/plans/packit-integration.fmf
+++ b/plans/packit-integration.fmf
@@ -1,0 +1,20 @@
+discover:
+    how: fmf
+    url: https://github.com/packit/packit
+    filter: tier:0 | tier:1
+
+adjust:
+  - when: "how == integration"
+    because: "provide latest python-specfile rpm when running locally"
+    prepare+:
+      - name: python3-specfile rpm
+        how: install
+        directory: noarch/
+
+  - when: "distro == rhel-9 or distro == centos-9 or distro == centos-stream-9"
+    because: "flexmock and deepdiff are not in EPEL 9: https://bugzilla.redhat.com/show_bug.cgi?id=2120251"
+    prepare+:
+      - how: install
+        package: python3-pip
+      - how: shell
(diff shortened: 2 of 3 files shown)


FrNecas:
I am also not completely sure what we want to do downstream with packit tests. Currently, `main` will be used to fetch the tests, I believe

FrNecas:
The tests fail both locally (try `tmt -c how=integration run -avvv plan -n packit-integration provision -h virtual -i fedora-36 login -s prepare` for testing with f36 build from `packit build`) and in Testing Farm, pretty cool.

nforro:
> Should we adjust this using `git_reference.py`?
I think that would make sense.

plans/packit-integration.fmf
```diff
@@ -0,0 +1,20 @@
...
+    url: https://github.com/packit/packit
+    filter: tier:0 | tier:1
+
+adjust:
+  - when: "how == integration"
+    because: "provide latest python-specfile rpm when running locally"
+    prepare+:
+      - name: python3-specfile rpm
+        how: install
+        directory: noarch/
```
nforro, line 12: So, when running locally, one is supposed to build a rpm and put it in `noarch` directory?

nforro: Ok, I see, `packit build locally` does that. I think we should document that somewhere (and probably also how to run other test plans).

nforro (approved):

nforro:
> I am also not completely sure what we want to do downstream with packit tests. Currently, `main` will be used to fetch the tests, I believe. Should we adjust this using `git_reference.py`?

Actually, won't the `git_reference.py` script overwrite `url` also in `packit-integration.fmf`?

plans/packit-integration.fmf
FrNecas: Yes, I was thinking what's the best place for this. In tmt, we used something like `<repo root>/tmp/RPMs`. But I found it quite convenient to just run `packit build locally` and then run the test without any need for moving things around. Or do you see a better place where to put RPMs to be installed when running tests locally?

nforro: No, I think it's fine, we just need to document it.

FrNecas: Done, I added a `CONTRIBUTING.md` describing this.

FrNecas:
@nforro PTAL

plans/git_reference.py
```diff
@@ -4,11 +4,23 @@
...
     )
+
+
+# Set discover of packit integration tests to the current main
+packit_node = tree.find("/plans/packit-integration")
+r = requests.get("https://api.github.com/repos/packit/packit/commits/main")
+if not r.ok:
+    raise SystemExit(1)
+with packit_node as data:
+    data["discover"]["ref"] = r.json()["sha"]
```
nforro, line 26: Since we are already calling git above, wouldn't it be better to call
`git ls-remote https://github.com/packit/packit main` to get the SHA?

nforro:
Nice job on `CONTRIBUTING.md` :+1:

FrNecas:
Hm, interesting, the centos-stream-9 test seems to still use the previous epel-9 build that had my intentional bug as per the logs... https://artifacts.dev.testing-farm.io/9433af99-2bcc-4ff0-8dc9-3061bbcf72a4/ this could be a Packit-service bug, I will try to investigate.

nforro (approved):

FrNecas:
The failing stream9 test is caused by a weird downgrade during Testing Farm test run to a previous build from this PR (a correct build artifact is installed at first but then overridden by an incorrect one). Testing Farm tracks this issue and plans to lock the version of artifacts so that this does not happen.

plans/packit-integration.fmf
```diff
@@ -0,0 +1,20 @@
+discover:
+    how: fmf
+    url: https://github.com/packit/packit
+    filter: tier:0 | tier:1
```
TomasTomecek, line 4: wow, this is so impressive; with 4 lines you get to run reverse dependency tests

PR #166, opened 2022-12-21
Fix normalizing name of a section

nforro:
Make only the first word (the actual name) lowercase.


nforro:
Though I'm wondering if the name `name` actually makes sense here. Wouldn't something like `id` be more fitting?

If we take for example `%files`, `%files doc` and `%files -f %{pyproject_files}`, at least the last case isn't really a name, it's a name and argument(s).

nforro:
Dropped in favor of #167.

PR #170, opened 2022-12-24
Add Section.is_script property

nforro:
Convenience property to determine if the content of a section is in fact a shell script.


TomasTomecek (approved):
LGTM though a small unit test case wouldn't hurt :)

nforro:
> LGTM though a small unit test case wouldn't hurt :)

Test added.

PR #173, opened 2022-12-27
Properly handle exceptions when re-parsing with dummy sources

### LINKED ISSUE #172: UnboundLocalError: cannot access local variable 'spec' where it is not associated with a value
Trying to parse `grub2.spec` from Fedora results in

```
    specfile = Specfile(filename, force_parse=True)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/specfile/specfile.py", line 57, in __init__
    self._parser.parse(str(self))
  File "/usr/lib/python3.11/site-packages/specfile/spec_parser.py", line 303, in parse
    self.spec, self.tainted = self._do_parse(content, extra_macros)
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/specfile/spec_parser.py", line 281, in _do_parse
    del spec
        ^^^^
UnboundLocalError: cannot access local variable 'spec' where it is not associated with a value
```

nforro:
Fixes #172.


TomasTomecek (commented):
would it make sense to include the excerpt of the grub2 spec file in tests to account for this use case?

nforro:
> would it make sense to include the excerpt of the grub2 spec file in tests to account for this use case?

I can make a test case for this.

specfile/spec_parser.py
```diff
@@ -250,19 +252,20 @@ def collect_included_sources(content):
...
-                        tainted = True
-                        # do a non-build parse again with dummy sources
-                        spec = get_rpm_spec(
-                            content, rpm.RPMSPEC_ANYARCH | rpm.RPMSPEC_FORCE
-                        )
-                        # spec.sources contains also previously collected
-                        # non empty sources (if any), remove them
-                        sources = {s for s, _, _ in spec.sources} - non_empty_sources
+                    if not dummy_sources:
+                        raise
```
mfocko, line 256: That made me a think for a moment :D

mfocko (commented):
+1 after the tests

nforro:
I've added a test case and I've also added support for `%{load:...}`, which is an equivalent of `%include ...`, but only for macro files (where macros are defined directly without `%global` or `%define`).

TomasTomecek (approved):
👏

PR #169, opened 2022-12-23
Format repr of all classes

nforro:
I've added a simple expression formatter that should follow `black` formatting style, and `@formatted` decorator to auto-format `__repr__()` in all classes.

Before:
```python
>>> spec.tags().content
Tags([Tag('Name', 'python-specfile', 'python-specfile', ':           ', Comments([], ['%if 0%{?rhel}
== 9', '# RHEL 9 is missing python-flexmock', '%bcond_with tests', '%else', '%bcond_without tests',
'%endif', '', '', '%global desc %{expand:', 'Python library for parsing and manipulating RPM spec
files.', 'Main focus is on modifying existing spec files, any change should result', 'in a minimal
diff.}', '', '']), '', ''), Tag('Version', '0.9.0', '0.9.0', ':        ', Comments([], []), '', ''),
Tag('Release', '1%{?dist}', '1.fc37', ':        ', Comments([], []), '', ''), Tag('Summary', 'A
library for parsing and manipulating RPM spec files', 'A library for parsing and manipulating RPM
spec files', ':        ', Comments([], ['']), '', ''), Tag('License', 'MIT', 'MIT', ':        ',
Comments([], []), '', ''), Tag('URL', 'https://github.com/packit/specfile',
'https://github.com/packit/specfile', ':            ', Comments([], []), '', ''), Tag('Source0',
'%{pypi_source specfile}',
'https://files.pythonhosted.org/packages/source/s/specfile/specfile-0.9.0.tar.gz', ':        ',
Comments([], ['']), '', ''), Tag('BuildArch', 'noarch', 'noarch', ':      ', Comments([], ['']), '',
''), Tag('BuildRequires', 'python%{python3_pkgversion}-devel', 'python3-devel', ':  ', Comments([],
['']), '', '')], ['', ''])

>>> spec.prep().content
Prep(PrepMacros([AutosetupMacro('%autosetup', MacroOptions([Token(<TokenType.DEFAULT: 1>, '-p1'),
Token(<TokenType.WHITESPACE: 2>, ' '), Token(<TokenType.DEFAULT: 1>, '-n'),
Token(<TokenType.WHITESPACE: 2>, ' '), Token(<TokenType.DEFAULT: 1>, 'specfile-%{version}')],
'a:b:cDn:TvNS:p:', {'n': '%{name}-%{version}', 'S': 'patch'}), ' ', '', '', [])], ['# Use packaged
[...]

PR #174, opened 2022-12-27
Support tuples in expression formatter

nforro:
A follow-up to #169. Needed because `Specfile.macros` is a list of tuples.


nforro:
recheck

TomasTomecek (approved):

PR #176, opened 2023-01-03
Make all classes consistently (deep)copyable

### LINKED ISSUE #55: Make all classes consistently (deep)copyable
Most classes do not implement `__copy__()`, and no class implements `__deepcopy__()` (though maybe it's not necessary). Some classes implement `copy()` instead. Make it consistent.

nforro:
Most classes don't need to implement anything and just work with both `copy()` and `deepcopy()`, only `SpecParser`, `SourcelistEntry` and `Sources` require special handling.

Classes inheriting from `collections.UserList` have to reimplement the `copy()` method, so it returns a shallow copy of the class, in a way similar to how `list.copy()` behaves.

Since `Specfile` can now be copied, it makes sense to make read-only properties writable, so one can adjust attributes of a copy.

It also makes more sense to parse the spec file before access to the underlying `rpm.spec` object or macros, rather than after 
updates of the content.

Fixes #55.

RELEASE NOTES BEGIN

All classes including `Specfile` itself can now be copied using the standard `copy()` and `deepcopy()` functions from `copy` module.

RELEASE NOTES END

### DIFF
--- a/specfile/changelog.py
+++ b/specfile/changelog.py
@@ -63,6 +63,15 @@ class ChangelogEntry:
             following_lines.copy() if following_lines is not None else []
         )
 
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, ChangelogEntry):
+            return NotImplemented
+        return (
+            self.header == other.header
+            and self.content == other.content
+            and self._following_lines == other._following_lines
+        )
+
     def __str__(self) -> str:
         return f"{self.header}\n" + "\n".join(self.content) + "\n"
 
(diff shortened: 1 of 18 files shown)


nforro:
I'm going to add unit tests for (deep)copying every class.

TomasTomecek (commented):
agreed with adding unit tests for the deepcopy functionality: that can prevent future headaches :)

TomasTomecek (approved):
very nice work with the tests

PR #177, opened 2023-01-05
Switch from setup.py to pyproject.toml

nforro:


.packit.yaml
```diff
@@ -30,15 +30,15 @@ actions:
...
     - bash -c "ls -1t ./fedora/*.tar.gz | head -n 1"
-  get-current-version: python3 setup.py --version
+  get-current-version: python3 -m setuptools_scm
   pre-sync:
     # FMF has to be installed on system where you are calling this tool.
     - python3 plans/git_reference.py
 
 srpm_build_deps:
-  - python3-pip # "python3 setup.py --version" needs it
+  #  - python3-pip # setuptools_scm needs it XXX: does it really?
```
TomasTomecek, line 41: interesting, I'd think that [build](https://github.com/pypa/build) needs to be here

pyproject.toml
```diff
@@ -0,0 +1,62 @@
+[build-system]
+requires = [
+    "setuptools>=61.0",
+    "setuptools_scm[toml]>=6.2",
+    "setuptools_scm_git_archive",
+]
+build-backend = "setuptools.build_meta"
```
TomasTomecek, line 7: ahaaa, so this is how you specify the build tool

TomasTomecek (approved):
if all builds pass

I'm curious about el9

.packit.yaml
nforro: Yep, I've just added it.

nforro:
So, EPEL 9 and even Fedora 36 are a problem. But I think we can keep `setup.cfg` and `setup.py`, even though it means duplicated metadata.

jpopelka:
https://stackoverflow.com/questions/62983756/what-is-pyproject-toml-file-for says that you actually still need `setup.py` (with just `import setuptools; setuptools.setup()`) for [editable installs](https://stackoverflow.com/questions/35064426/when-would-the-e-editable-option-be-useful-with-pip-install).

nforro:
> https://stackoverflow.com/questions/62983756/what-is-pyproject-toml-file-for says that you actually still need `setup.py` (with just `import setuptools; setuptools.setup()`) for [editable installs](https://stackoverflow.com/questions/35064426/when-would-the-e-editable-option-be-useful-with-pip-install).

Actually: https://stackoverflow.com/a/68939858
> As of August 2022, Setuptools and Pip now fully support PEP 660, and therefore it is now possible to perform an editable installation with only pyproject.toml.

nforro:
So, this sucks. `pyproject-rpm-macros` and `pip` in EL9 and Fedora 36 support `pyproject.toml`, but the version of `setuptools` is too old and unable to get metadata from it. And the `pyproject.toml` file must not exist for a fallback to `setup.py`.

gotmax23:
> So, this sucks. `pyproject-rpm-macros` and `pip` in EL9 and Fedora 36 support `pyproject.toml`, but the version of `setuptools` is too old and unable to get metadata from it. And the `pyproject.toml` file must not exist for a fallback to `setup.py`.

Not necessarily.  You should be able to keep the metadata in setup.cfg, get rid of setup.py, get rid of setup_requires, and add a pyproject.toml with

``` toml
[build-system]
requires = [
    "setuptools>=45",
    "setuptools_scm[toml]>=6.2"
]
build-backend = "setuptools.build_meta"

[tool.setuptools_scm]
version_scheme = "post-release"
```

and maintain compatibility with EL 9 and current Fedora releases.

nforro:
Nice, let me try that.

nforro:
Hm, EPEL9 has `setuptools_scm` 6.0.1, so `python3 -m setuptools_scm` doesn't really work, but the wheel build is somehow able to determine the correct version.

Thanks @gotmax23!

gotmax23:
> Hm, EPEL9 has `setuptools_scm` 6.0.1, so `python3 -m setuptools_scm` doesn't really work, but the wheel build is somehow able to determine the correct version.

You should be able to work around this by replacing `python3 -m setuptools_scm` with `python -c 'from setuptools_scm import get_version; print(get_version(version_scheme="post-release"))'`.

> Thanks @gotmax23!

Sure :).

gotmax23:
> You should be able to work around this by replacing `python3 -m setuptools_scm` with `python -c 'from setuptools_scm import get_version; print(get_version(version_scheme="post-release"))'`.

Before https://github.com/pypa/setuptools_scm/commit/fc3dd17b332a76986da00d03acbb80da760bebb8 (6.4.0), `python -m setuptools_scm` didn't consider the version scheme specified in the config file.

nforro:
Good to know. However, it shouldn't be an issue here, since the `get-current-version` action (correct me if I'm wrong) runs on Fedora 37. As long as the build on EPEL9 works correctly, we are ok.

TomasTomecek (approved):
wow, impressive!

@nforro and @gotmax23, you are heroes!

PR #181, opened 2023-01-08
Add evr argument to Specfile.add_changelog_entry

gotmax23:
This allows adding a changelog entry with an evr value that's different than the current specfile's value. This makes it easier to reconstruct a specfile's %changelog based on another source using the higher level interface.

TODO:

RELEASE NOTES BEGIN

Add evr argument to Specfile.add_changelog_entry. This allows adding a changelog entry with an evr value that's different than the current specfile's value. This makes it easier to reconstruct a specfile's %changelog based on another source using the higher level interface.

RELEASE NOTES END

### DIFF
--- a/specfile/specfile.py
+++ b/specfile/specfile.py
@@ -393,6 +393,7 @@ class Specfile:
         author: Optional[str] = None,
         email: Optional[str] = None,
         timestamp: Optional[Union[datetime.date, datetime.datetime]] = None,
+        evr: Optional[str] = None,
     ) -> None:
         """
         Adds a new %changelog entry. Does nothing if there is no %changelog section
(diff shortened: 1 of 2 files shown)


nforro (approved):
Thanks!

specfile/specfile.py
```diff
@@ -407,15 +408,19 @@ def add_changelog_entry(
             email: E-mail of the author.
             timestamp: Timestamp of the entry.
               Supply `datetime` rather than `date` for extended format.
+            evr: Override the evr part of the changelog.
```
nforro, line 411: Maybe it's better to capitalize the abbreviation:
```suggestion
            evr: Override the EVR part of the changelog.
```

gotmax23:
I rebased on top of `main` and applied your suggested change with one small modification

gotmax23:
Thanks for reviewing and merging!

PR #189, opened 2023-01-17
Fix expansion of tag values

nforro:
Retrieving expanded values of tags from parsed sections doesn't make sense because it's impossible to undoubtedly match an unparsed tag to a parsed one. The same goes for validity of a tag.

Expand tag values on demand using macro expansion and remove the `valid` property.

### DIFF
--- a/specfile/tags.py
+++ b/specfile/tags.py
@@ -5,12 +5,27 @@ import collections
 import copy
 import itertools
 import re
-from typing import Any, Iterable, List, Optional, SupportsIndex, Union, cast, overload
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    SupportsIndex,
+    Union,
+    cast,
+    overload,
+)
 
 from specfile.constants import TAG_NAMES, TAGS_WITH_ARG
 from specfile.formatter import formatted
+from specfile.macros import Macros
 from specfile.sections import Section
 from specfile.utils import split_conditional_macro_expansion
+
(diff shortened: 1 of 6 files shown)


specfile/tags.py
```diff
@@ -195,11 +209,11 @@ def __init__(
         self,
         name: str,
         value: str,
-        expanded_value: Optional[str],
         separator: str,
         comments: Comments,
         prefix: Optional[str] = None,
         suffix: Optional[str] = None,
+        context: Optional["Specfile"] = None,
```
jpopelka, line 216: There are missing attributes in the docstring.

specfile/tags.py
```diff
@@ -225,19 +239,18 @@ def __init__(
...
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, Tag):
             return NotImplemented
         return (
             self.name == other.name
             and self.value == other.value
-            and self._expanded_value == other._expanded_value
             and self._separator == other._separator
             and self.comments == other.comments
             and self._prefix == other._prefix
```
jpopelka, line 256: The context doesn't need to be compared?

nforro: The idea here (and in `SourcelistEntry.__eq__()`) was to be able to compare tags between different `Specfile` objects. But thinking about it, maybe it doesn't make a lot of sense. What do you think?

Anyway, this made me realize that the `_context` attribute needs to be be excluded from deepcopying, I'll fix that.

nforro:
Packit test failures are probably flakes, definitely unrelated to changes in this PR.

PR #207, opened 2023-02-21
Get rid of rpm-py-installer

### LINKED ISSUE #187: rpm-py-installer deprecation warnings
rpm-py-installer is unable to build wheels and relies on `setup.py install` and eggs. Currently, pip prints obnoxious warnings when the wheel build fails and it falls back to `setup.py install`. This fallback will be removed altogether in pip 23.1. I think it's time to remove this dependency from the pypi package and find another solution. Potential solutions:

1. Keep the dependency. specfile will become uninstallable with newer pip versions.
2. Remove the dependency entirely and advise consumers to use the python3-specfile RPM package or to install python3-rpm and use `pip install --user specfile` or venvs with `--system-site-packages`. `ImportError`s will be raised when `rpm` isn't available.
3. Port the existing rpm-py-installer code (if feasible).
4. Create a new implementation of rpm-py-installer. This could be a shim module that tries to import `rpm` from the system site-packages (a la [pycontribs/selinux](https://github.com/pycontribs/selinux/)).

nforro:
Fixes #187.

RELEASE NOTES BEGIN

Specfile no longer depends on rpm-py-installer, it now depends directly on rpm.

RELEASE NOTES END

### DIFF
--- a/fedora/python-specfile.spec
+++ b/fedora/python-specfile.spec
@@ -41,8 +41,6 @@ Summary:        %{summary}
 
 %prep
 %autosetup -p1 -n specfile-%{version}
-# Use packaged RPM python bindings downstream
-sed -i 's/rpm-py-installer/rpm/' setup.cfg
 
 
 %generate_buildrequires
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,7 @@ keywords =
 [options]
 packages = find:
 install_requires =
-    rpm-py-installer
+    rpm
 python_requires = >=3.9
 include_package_data = True
 


TomasTomecek (approved):
sweet!

but what's wrong with the tests?

TomasTomecek:
I can see similar build failures: https://github.com/packit/packit/issues/1855

nforro:
> but what's wrong with the tests?

I'm not sure, but I don't think the failures are related to this change.

nforro:
recheck

TomasTomecek (approved):
let's go

nforro:
recheck

nforro:
> There is a packit test that seems to fail consistently: `test_local_build_from_srpm`. I'm not sure how it could be related, but I want to take a closer look before I merge this PR.

It happens on every specfile PR, so it's definitely unrelated, let's not block this further.

PR #213, opened 2023-03-02
Enable packit-dev repo for the packit-integration test plan

### LINKED ISSUE #211: Packit integration test plan doesn't work properly
packit test suite is fetched from the main branch of packit git repository, but it runs with packit installed from (stable) RPM. That is asking for trouble - for instance a test that doesn't have the corresponding implementation in the stable release yet can be run and fail.

For downstream there is `git_reference.py` script that hardcodes the corresponding git ref to test plans, perhaps we could use something similar here?

nforro:
Fixes #211.

### DIFF
--- a/plans/packit-integration.fmf
+++ b/plans/packit-integration.fmf
@@ -3,6 +3,13 @@ discover:
     url: https://github.com/packit/packit
     filter: tier:0 | tier:1
 
+prepare:
+  - how: install
+    copr: packit/packit-dev
+  # make sure the Copr repo has higher priority than TF Tag Repository
+  - how: shell
+    script: dnf -y config-manager --save --setopt="*:packit:packit-dev.priority=5"
+
 adjust:
   - when: "how == integration"
     because: "provide latest python-specfile rpm when running locally"


plans/packit-integration.fmf
```diff
@@ -3,6 +3,10 @@ discover:
     url: https://github.com/packit/packit
     filter: tier:0 | tier:1
 
+prepare:
+    how: install
```
FrNecas, line 7: Based on the error in EPEL, this has to be a list because of the adjust. In the adjust, we use `    prepare+:
`, note the plus - it expects lists to be merged. So hopefully

```
prepare:
  - how: install
    copr: ...
```

should remove the error...

FrNecas:
But I am not sure why there are so many failing tests on rawhide. Weird traceback...

nforro:
> But I am not sure why there are so many failing tests on rawhide. Weird traceback...

That should be already solved: https://github.com/packit/packit/pull/1858
But it still fails, so the packit version installed is not the one from packit-dev :disappointed:

FrNecas:
Let's see what happens now after your last push. I am a bit suspicious that it may be related to Testing Farm not locking dependencies so even if the correct version is installed, it is downgraded somewhere down the line.

nforro:
According to the [log](https://artifacts.dev.testing-farm.io/df80a5a6-b46f-4af5-925f-12bac55f178b/work-packit-integration3cia80rf/log.txt) packit is installed from `testing-farm-tag-repository` despite the Copr repo being enabled (you can see that it's actually refreshing its metadata: `Copr repo for packit-dev owned by packit         48 kB/s | 9.9 kB     00:00`).

FrNecas:
Hm, it looks like enabling copr repositories in testing farm does not work as I would expect. I will try to ask around :)

nforro:
Explicitly listing packages to install doesn't make a difference either. The `testing-farm-tag-repository` seems to be configured to have absolute priority. I wonder if we could do something about it in `tf_post_install_script` :thinking:

nforro:
Thanks, that's what I had in mind (except that `tf_post_install_script` is probably not helpful here).
Looking at the tag repo [config template](https://gitlab.com/testing-farm/infrastructure/-/blob/6ad82e9547b48765610e7e14c07c0bdfcf45ec4d/ranch/public/citool-config/guest-setup/pre-artifact-installation/templates/tag.repo.j2#L6), it seems it would be enough if we could add the Copr repo with lower proprity than 9. That way everything else would still be installed from the tag repo and nothing but packit (and its deps from the Copr repo) would be affected. Question is, how to do that :slightly_smiling_face:

nforro:
So, it still doesn't work, even though the priorities are set correctly:
```
out: /etc/yum.repos.d/tag-repository.repo:priority=9
out: /etc/yum.repos.d/_copr:copr.fedorainfracloud.org:packit:packit-dev.repo:priority=5
```

nforro:
The problem is in the Copr repo itself, there are no packit builds for F38 and rawhide.

FrNecas:
Oh, looks like we build just for fedora stable and epel: https://github.com/packit/packit/blob/main/.packit.yaml#L78 . If we want to make this work, we probably need to build for rawhide too? Or are there any alternatives?

nforro:
See https://github.com/packit/packit/pull/1867. There was also an issue with F38 not being automatically enabled because rawhide was disabled.

nforro:
Ok, so now the proper packit version is installed, let me try without adjusting the priority, maybe enabling the Copr repo is enough.

nforro:
> let me try without adjusting the priority, maybe enabling the Copr repo is enough.

Nope, it isn't.

nforro:
Ok, everything should be in order now, even though it doesn't look like it :slightly_smiling_face: 
F38 packit build still isn't in the packit-dev repo, and won't be until this is deployed: https://github.com/packit/packit/pull/1863, so we can ignore it for now.
As for the rawhide failures, I assume they've been there for a while and went unnoticed so far because packit [doesn't run](https://github.com/packit/packit/blob/a9640d26bb7ff09f29d26591b1152aa634e98f7a/.packit.yaml#L55) TF tests in rawhide.

FrNecas (approved):
Looks good, thanks for debugging the problem :) Those failures on rawhide seem fixable quite easily, a few tests need to be changed (some mock values used result in an error).

nforro:
Thanks.

> Those failures on rawhide seem fixable quite easily, a few tests need to be changed (some mock values used result in an error).

Yes, I'm working on that right now :slightly_smiling_face:

PR #225, opened 2023-05-04
ci: add test coverage to pytest

mfocko:


nforro:
Do we want to enable this in TF as well?
https://github.com/packit/specfile/blob/7a4e2cace13bd0db5c5ae777ca0d8072fec0c638/tests/full.fmf#L16

nforro (approved):

nforro:
recheck

PR #235, opened 2023-05-31
add test that tries to parse all Fedora packages

xsuchy:
The run takes about 90 minutes on my notebook.

- [x] Write new tests or update the old ones to cover new functionality.
- [-] Update doc-strings where appropriate.
- [-] Update or write new documentation in `packit/packit.dev`.

PR #237, opened 2023-06-11
Add functional test to parse all Fedora spec files

nforro:
Inspired by #235.


mfocko:
```
================ 2812 failed, 20192 passed in 968.78s (0:16:08) ================
```

Are you planning on fixing it? :eyes:

nforro:
It will go down to about 10 failures once I figure out and install all the macro packages that are necessary for parsing :slightly_smiling_face:

mfocko:
Ah sure, it seemed like there are some “clusters”, I think Go was one of those?

nforro:
Yes. Let's see now.

nforro:
Ok, https://github.com/rpm-software-management/rpm/commit/b3adc8ce62517cd764a589d2bdccbf97dadb915a explains the higher number of failures on rawhide. As for the rest, some of them are bugs (and for some of those there are already issues reported, thanks @xsuchy) and a few of them are packaging errors, I'll try to find a way to differentiate those.

tests/functional.fmf
```diff
@@ -0,0 +1,37 @@
...
+  - ocaml-srpm-macros
+  - openblas-srpm-macros
+  - package-notes-srpm-macros
+  - perl-srpm-macros
+  - pyproject-srpm-macros
+  - python-srpm-macros
+  - qt5-srpm-macros
+  - qt6-srpm-macros
+  - rust-srpm-macros
+  - typelib-srpm-macros
```
xsuchy, line 30: Does the fmf accepts `*-srpm-macros`?

nforro: Let me try, dnf seems to support it, so it could work.

PR #249, opened 2023-07-12
Use Centos Stream 9 base image

jpopelka:
Related to packit/deployment#410


nforro:
> How do you run the tests locally?

Since all tests are self-contained I've always run them with pytest directly.

PR #253, opened 2023-07-13
Fix tokenization of option strings

nforro:
Adjust the tokenization algorithm to take into account quotes contained within macro expansions and other expressions.

Notes for reviewers:

The tokenization works as before, but the option string is first parsed into nodes and everything that's not a string literal is treated as a single token.

RELEASE NOTES BEGIN

Fixed a bug in processing options of `%prep` macros. For instance, when a quoted string appeared inside an expression expansion, it could lead to improper parsing, rendering the spec file invalid after accessing the options.

RELEASE NOTES END


lbarcziova (approved):
lgtm :+1:

PR #261, opened 2023-07-30
Fix Packit config

nforro:
It turns out it's not possible to define certain keys at the top level, because they are not propagated to packages.


nforro:
recheck

majamassarini (approved):

lachmanfrantisek (approved):

nforro:
Merging manually.

PR #269, opened 2023-08-04
make sure %license is SPDX compatible

TomasTomecek:
```
$ askalono crawl LICENSE
LICENSE
License: MIT (original text)
Score: 1.000
```

https://fedoraproject.org/wiki/Changes/SPDX_Licenses_Phase_1

The changelog should propagate to dist-git as the SPDX migration
initiative tracks 'spdx' string in changelogs.

I'm at the SPDX hackfest right now and the guidance I got about the
LICENSE Copyright line was:
1. it's meaningless and a waste of time to polish
2. only the "copyright" word has a meaning
3. the year is meaningless, hence I dropped it
4. the team name & authors is relevant so I updated it to match 100%
   what we have in the source files

RELEASE NOTES BEGIN

Specfile's license in RPM spec file is now confirmed to be SPDX compatible.

RELEASE NOTES END

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2021 Red Hat
+Copyright Contributors to the Packit project.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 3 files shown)


nforro (approved):

TomasTomecek:
right, the problem with statuses is the one Maja described on standup

lbarcziova:
this should have been already done in specfile (https://src.fedoraproject.org/rpms/python-specfile/blob/rawhide/f/python-specfile.spec#_103), but doesn't hurt to have it twice probably :D

PR #271, opened 2023-08-09
Handle failures when creating dummy sources

nforro:
This makes it possible to parse a spec file on a read-only filesystem for instance. At the cost of incorrectly parsed `%prep` in case some sources are missing.

Related to https://github.com/fedora-copr/copr/issues/2839.


specfile/spec_parser.py
```diff
@@ -99,6 +99,20 @@ def _make_dummy_sources(
         Yields:
             List of paths to each created dummy source.
         """
+
+        def write_magic(path, magic):
```
lbarcziova, line 103: docstrings with a little bit of explanation could be helpful here

lbarcziova (approved):

specfile/spec_parser.py
nforro: I've added a comment.

lbarcziova: thanks!

Issue #231, opened 2023-05-22
Implement parsing of conditions


nforro:
Statements such as `%if`, `%else` and `%endif` are currently treated as regular section content.
That can cause issues, especially with conditions spanning across multiple sections - if a section is deleted, modified or moved, it can lead to syntax errors as the condition block could be broken apart.
Figure out how to deal with this.

lachmanfrantisek:
Still relevant. Let us know if you need this (either in Specfile or for Packit).

xsuchy:
I am hitting this when parsing amavis.spec. I get this TB
```
Traceback (most recent call last):
  File "/home/msuchy/projects/license-validate/./convert-spec-universal.py", line 65, in <module>
    with specfile.tags(section) as tags:
         ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/msuchy/.local/lib/python3.12/site-packages/specfile/context_management.py", line 133, in __call__
    self.values[key] = next(self.generators[key])
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/msuchy/.local/lib/python3.12/site-packages/specfile/specfile.py", line 292, in tags
    tags = Tags.parse(section, context=self)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/msuchy/.local/lib/python3.12/site-packages/specfile/tags.py", line 497, in parse
    lines = process_conditions(list(section), macro_definitions, context)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/msuchy/.local/lib/python3.12/site-packages/specfile/conditions.py", line 133, in process_conditions
    result.append((line, branches[-2]))
                         ~~~~~~~~^^^^
IndexError: list index out of range
```

xsuchy:
Parsing of `pg_auto_failover.spec` from Fedora is hitting this issue as well. :(

xsuchy:
Parsing of python3.14 is affected by this too.

xsuchy:
This is complete list of packages from Fedora affected by this issue:
```
amavis.spec
autoconf.spec
dogtag-pki.spec
frysk.spec
gdb.spec
gnutls.spec
grub2.spec
libpst.spec
make.spec
mariadb10.11.spec
mariadb11.8.spec
mysql8.0.spec
mysql8.4.spec
nasm.spec
[...]

PR #276, opened 2023-08-20
Implement processing of conditions

nforro:
This is not a full support for conditions, i.e. it doesn't re-solve https://github.com/packit/specfile/issues/231 where an entire section can be conditionalized (however, the new `process_conditions()` function is generic enough to be used as a base for the resolution of that issue), nevertheless I think it's quite an improvement.

Note that macro definitions have to be parsed twice, because a macro definition can contain a condition (see https://src.fedoraproject.org/rpms/kernel/blob/rawhide/f/kernel.spec for examples), and vice versa, a condition can encapsulate a macro definition.

Macro definitions and tags gained a `valid` attribute that can be used to determine if that particular macro definition/tag is valid and can affect other entities in the spec file or if it would be ignored when parsing the spec file with RPM. This will be used to fix `Specfile.update_value()` and `Specfile.update_tag()`, but it could be beneficial for other use cases as well.

Related to https://github.com/packit/packit/issues/2033.

RELEASE NOTES BEGIN

Macro definitions and tags gained a new `valid` attribute. A macro definition/tag is considered valid if it doesn't appear in a false branch of any condition appearing in the spec file.

RELEASE NOTES END

### DIFF
--- a/specfile/conditions.py
+++ b/specfile/conditions.py
@@ -1,0 +1,137 @@
+# Copyright Contributors to the Packit project.
+# SPDX-License-Identifier: MIT
+
+import re
+from typing import TYPE_CHECKING, List, Optional, Tuple
+
+from specfile.exceptions import RPMException
+from specfile.macros import Macros
+
+if TYPE_CHECKING:
+    from specfile.macro_definitions import MacroDefinitions
+    from specfile.specfile import Specfile
+
+
+def resolve_expression(
+    keyword: str, expression: str, context: Optional["Specfile"] = None
+) -> bool:
+    """
+    Resolves a RPM expression.
+
+    Args:
+        keyword: Condition keyword, e.g. `%if` or `%ifarch`.
+        expression: Expression string or a whitespace-delimited list
+          of arches/OSes in case keyword is a variant of `%ifarch`/`%ifos`.
+        context: `Specfile` instance that defines the context for macro expansions.
+
+    Returns:
+        Resolved expression as a boolean value.
+    """
+
+    def expand(s):
+        if not context:
+            return Macros.expand(s)
+        result = context.expand(s, skip_parsing=getattr(expand, "skip_parsing", False))
+        # parse only once
+        expand.skip_parsing = True
+        return result
+
+    if keyword in ("%if", "%elif"):
+        try:
+            result = expand(f"%{{expr:{expression}}}")
+        except RPMException:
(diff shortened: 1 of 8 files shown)


lachmanfrantisek:
> Note that macro definitions have to be parsed twice

Are there any performance concerns? (E.g. in Copr)

nforro:
> Are there any performance concerns? (E.g. in Copr)

Well, it will make things slower, but it shouldn't be noticeable. Parsing `texlive.spec` takes roughly 100ms more.

specfile/conditions.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def expand(s):
+        if context:
+            result = context.expand(
+                s, skip_parsing=getattr(expand, "skip_parsing", False)
+            )
+            # parse only once
+            expand.skip_parsing = True
+            return result
+        return Macros.expand(s)
```
lachmanfrantisek, line 39: I am having trouble remembering contexts..;)

```suggestion
        if not context:
            return Macros.expand(s)
        
        result = context.expand(
             s, skip_parsing=getattr(expand, "skip_parsing", False)
        )
        # parse only once
        expand.skip_parsing = True
        return result
```

specfile/conditions.py
```diff
@@ -0,0 +1,141 @@
...
+    branches = [True]
+    indexed_lines = list(enumerate(lines))
+    while indexed_lines:
+        index, line = indexed_lines.pop(0)
+        # ignore conditions inside macro definition body
+        if any(index in r for r in excluded_lines):
+            result.append((line, branches[-1]))
+            continue
+        m = condition_regex.match(line)
+        if m:
```
lachmanfrantisek, line 112: Maybe a time for walrus operator?..;-D

lachmanfrantisek: I would also try to use the same trick as above (=reverse the condition and return early) to avoid ging one level deeper to the right.

lachmanfrantisek (approved):
Just wow, this really shows what it means to parse the specfile..;-D

I am only not sure if I am able to review this..;) So, just a few code-style nit picks.

specfile/conditions.py
nforro: > Maybe a time for walrus operator?..;-D

Not supported in Python 3.6.

nforro:
recheck

PR #280, opened 2023-08-30
Fix Specfile.update_value()

nforro:
Related to https://github.com/packit/packit/issues/2033.

### DIFF
--- a/specfile/macro_definitions.py
+++ b/specfile/macro_definitions.py
@@ -185,13 +185,14 @@ class MacroDefinitions(collections.UserList):
     def copy(self) -> "MacroDefinitions":
         return copy.copy(self)
 
-    def get(self, name: str) -> MacroDefinition:
-        return self.data[self.find(name)]
+    def get(self, name: str, position: Optional[int] = None) -> MacroDefinition:
+        return self.data[self.find(name, position)]
 
-    def find(self, name: str) -> int:
+    def find(self, name: str, position: Optional[int] = None) -> int:
         for i, macro_definition in enumerate(self.data):
             if macro_definition.name == name:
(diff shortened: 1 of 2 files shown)


nforro:
Good point, let me extend [test_update_tag](https://github.com/packit/specfile/blob/5781f2fd3fc0f383908281d3fe9ae3b21a302c0a/tests/integration/test_specfile.py#L354).

nforro:
It turns out RPM version in EL8 doesn't support the `%expr` macro and expression expansion doesn't work either, so it's not possible to evaluate an expression in a condition without reimplementing the entire expression solver.

I'm going to mark the `test_update_tag` as expected to fail with RPM older than 4.16 and mention that in the documentation.

nforro:
@lbarcziova Do you think I should also adjust [the code](https://github.com/packit/specfile/blob/5781f2fd3fc0f383908281d3fe9ae3b21a302c0a/specfile/conditions.py#L40-L47) to raise an exception when the evaluation doesn't work? The current implementation makes it silently pass, with every condition expression evaluated as `True`.

nforro:
> @lbarcziova Do you think I should also adjust [the code](https://github.com/packit/specfile/blob/5781f2fd3fc0f383908281d3fe9ae3b21a302c0a/specfile/conditions.py#L40-L47) to raise an exception when the evaluation doesn't work? The current implementation makes it silently pass, with every condition expression evaluated as `True`.

Actually, to follow the current logic, it makes more sense to just resolve is as `False`.

lbarcziova (approved):
> Actually, to stick with the current logic, it makes more sense to just resolve it as False.

+1

PR #286, opened 2023-09-18
Make modifiable entity groups greedy

nforro:
This effectively makes the first group most greedy, rather than the last. So, with the following in a spec file:

```spec
%global apiver 2.15
%global patchver 3

Version: %{apiver}.%{patchver}
```

calling `spec.update_tag("Version", "2.16.0")`, the `apiver` macro gets assigned the bigger portion of the version string, resulting in:

```spec
%global apiver 2.16
%global patchver 0
```

rather than:

```spec
%global apiver 2
%global patchver 16.0
```

There are cases (in Fedora spec files) where the previous logic makes more sense, but only a handful of them, and none of them using packit.
With the current implementation, it's one or the other, and I believe this makes more sense.

Related to https://github.com/packit/packit/issues/2033

### DIFF
--- a/specfile/value_parser.py
+++ b/specfile/value_parser.py
@@ -425,7 +425,7 @@ class ValueParser:
                     # this group name already exists, make a backreference
                     regex += f"(?P={group})"
                 else:
-                    regex += f"(?P<{group}>.+?)"
+                    regex += f"(?P<{group}>.+)"
                     named_groups.append(group)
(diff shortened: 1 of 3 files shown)


lachmanfrantisek (approved):
Can you please add a new test case for this?

lachmanfrantisek (approved):

PR #288, opened 2023-09-21
Stop using deprecated post-release versioning scheme

nforro:


pyproject.toml
```diff
@@ -7,4 +7,4 @@ requires = [
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools_scm]
-version_scheme = "post-release"
+version_scheme = "no-guess-dev"
```
lachmanfrantisek, line 10: Would it be possible to mention why we are using this schema? I always need to google it when I see it..;)

nforro: @lachmanfrantisek does this work? :slightly_smiling_face: 

https://github.com/packit/specfile/blob/446a315ca546b924627e4c89e19a4f654ff8609a/pyproject.toml#L10-L16

lachmanfrantisek: Sure, thanks a lot!

lachmanfrantisek (approved):

PR #295, opened 2023-10-25
Add valid property to sources

nforro:
Related to https://github.com/packit/packit/issues/2129.

RELEASE NOTES BEGIN

Sources now have a `valid` property that indicates whether a source is valid in the current context, meaning it is not present in a false branch of any condition.

RELEASE NOTES END

### DIFF
--- a/specfile/sourcelist.py
+++ b/specfile/sourcelist.py
@@ -5,7 +5,9 @@ import collections
 import copy
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, overload
 
+from specfile.conditions import process_conditions
 from specfile.formatter import formatted
+from specfile.macro_definitions import MacroDefinitions
 from specfile.macros import Macros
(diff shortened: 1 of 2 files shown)


gotmax23:
Note that conditionalizing Sources or Patches is a code smell and can have negative consequences (see e.g., https://docs.fedoraproject.org/en-US/packaging-guidelines/SourceURL/#_do_not_conditionalize_sources), but I guess it's still valid to implement this...

lbarcziova (approved):
:+1:

lachmanfrantisek:
> Note that conditionalizing Sources or Patches is a code smell and can have negative consequences (see e.g., https://docs.fedoraproject.org/en-US/packaging-guidelines/SourceURL/#_do_not_conditionalize_sources), but I guess it's still valid to implement this...

Maybe, we can warn the user and link this paragraph once we find such a case..;)

PR #298, opened 2023-11-02
Allow to comment out or uncomment a macro definition

nforro:
Necessary for Packit to be able to comment out or uncomment a pre-release suffix macro definition ([example](https://src.fedoraproject.org/rpms/golang/blob/5ccb908322500645653a1367f8f9bcded7af5574/f/golang.spec#_99)).

Related to https://github.com/packit/packit/issues/2013.

RELEASE NOTES BEGIN

Macro definitions gained a new `commented_out` property indicating that a macro definition is commented out. Another new property, `comment_out_style`, determines if it is achieved by using a `%dnl` (discard next line) directive (e.g. `%dnl %global prerelease beta2`) or by replacing the starting `%` with `#` (e.g. `#global prerelease beta2`).

RELEASE NOTES END

### DIFF
--- a/specfile/macro_definitions.py
+++ b/specfile/macro_definitions.py
@@ -4,6 +4,7 @@
 import collections
 import copy
 import re
+from enum import Enum, auto
 from typing import TYPE_CHECKING, List, Optional, Tuple, Union, overload
 
 from specfile.conditions import process_conditions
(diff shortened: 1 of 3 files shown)


gotmax23:
Why not use the `%dnl` macro which doesn't require escaping anything to comment out lines? (Note that this isn't available on older RPM versions.)

nforro:
> Why not use the `%dnl` macro which doesn't require escaping anything to comment out lines? (Note that this isn't available on older RPM versions.)

Mainly because nobody uses `%dnl` to switch from a pre-release, whereas commenting out seems to be widely used in Fedora (see for example https://src.fedoraproject.org/rpms/cmake/c/4e11520f2d913bce471f53156dfe21e77bee8215). I don't know if this "convention" is documented somewhere, but a lot of packages seem to have adopted it and I don't think we can just change it.

gotmax23:
Well, if you're adding a commented_out property, `%dnl %global ...` and `%dnl %define ...` should also be considered.

nforro:
> Well, if you're adding a commented_out property, `%dnl %global ...` and `%dnl %define ...` should also be considered.

Ok, I suppose it would make sense to add another property, let's say `disabled`, for `%dnl`. Or would you suggest a better name?

gotmax23:
Maybe have one `commented_out` attribute and another attribute that stores whether the comment character is `#` or `%dnl`? I don't personally see the value in differentiating the two comment styles with separate attributes, but if you think differently, I think `disabled` is a good name.

lachmanfrantisek (approved):
I like the approach with multiple styles.

PR #299, opened 2023-11-03
Add type arguments to UserList subclasses

gotmax23:
By having these classes inherit from the generics in the typing module, type checkers will be able to determine the return types of the dunder methods inherited from UserList.

RELEASE NOTES BEGIN

Improve type annotations for UserList subclasses

RELEASE NOTES END


nforro (approved):
Thanks!

PR #317, opened 2023-11-18
Implement updating version with support for pre-releases

nforro:
Moved the implementation from https://github.com/packit/packit/pull/2149 here and improved `update_value()`.

Related to https://github.com/packit/packit/issues/2013.

Merge before https://github.com/packit/packit/pull/2149.

RELEASE NOTES BEGIN

There is a new method, `Specfile.update_version()`, that allows updating spec file version even if it is a pre-release.

RELEASE NOTES END


specfile/specfile.py
```diff
@@ -826,3 +863,65 @@ def update_tag(
...
+              To be commented out or uncommented accordingly.
+            comment_out_style: Whether to use `%dnl` macro or swap the leading '%'
+              with '#' to comment out `prerelease_suffix_macro`. Defaults to `%dnl`.
+
+        Raises:
+            SpecfileException if `prerelease_suffix_pattern` is invalid.
+        """
+
+        def handle_prerelease(version):
+            def update_macro(prerelease_detected):
```
lachmanfrantisek, line 902: Would it be hard to factor these out? The closures are quite handy but having multiple levels makes it a bit too hard to read (at least) for me..;)

lachmanfrantisek (approved):
Maybe it's just Monday morning but I had a hard time going through the changes..;) But it is technical work and you have it nicely covered by the tests so :+1: 

Thanks Nikola!

specfile/specfile.py
nforro: I could make `update_macro()` be on the same level as `handle_prerelease()`, would that help?

nforro:
EL9 failures are not related to changes in this PR, I'll investigate and try to fix them in another PR not to block this one.

PR #340, opened 2024-01-08
Fix `update_value()`

nforro:
The implementation didn't account for a case when a macro definition needed to be commented out to succesfully set the new value but it was commented out already. This commit fixes that.


specfile/specfile.py
```diff
@@ -783,12 +783,16 @@ def update(value, requested_value, position):
...
-            m = regex.match(requested_value)
-            if m:
+            # in case the value doesn't match after trying with flippable entities
+            # do a second pass without them
+            for flippable_ents in (flippable_entities, set()):
+                regex, template, entities_to_flip = ValueParser.construct_regex(
+                    value, modifiable_entities, flippable_ents, context=self
+                )
+                m = regex.match(requested_value)
+                if not m:
```
lachmanfrantisek, line 794: I am wondering if walrus can help here..;)

nforro: We can't use that, specfile still has to support Python 3.6.

PR #343, opened 2024-01-12
Generate API documentation from docstrings

nforro:


majamassarini (approved):
Maybe I would add a comment to the EscapeBracketsProcessor because I am not really sure to understand why you need it 😅🙏

lbarcziova (approved):
nice!

PR #350, opened 2024-02-20
Ci/improvements

mfocko:


.github/workflows/check-release-notes.yml
```diff
@@ -0,0 +1,15 @@
...
+  pull_request:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+jobs:
+  check_release_notes:
+    name: Notes are either written, or there are none
+    uses: packit/.github/.github/workflows/check-release-notes.yml@main
```
nforro, line 13: Shouldn't this be:

```suggestion
    uses: packit/.github/workflows/check-release-notes.yml@main
```

mfocko: It used to be that, but it screws up the path deduction, cause you can have the action in the repo, but also reuse it from other of your repos… so I had to move it there :/

It's this path:
https://github.com/packit/.github/blob/main/.github/workflows/check-release-notes.yml

I probably could move it somewhere else? Or have a separate repo, this seems like reasonable option right now :shrug:

nforro: Oh, I see. A repo name starting with a dot is a bit confusing :slightly_smiling_face:

PR #356, opened 2024-03-19
Fix parsing of `%prep` macros

nforro:
Parsing of `%prep` macros was incorrect. For instance `%patches` macro when used in `%prep` section was wrongly parsed as `%patch` macro. This commit fixes that.

RELEASE NOTES BEGIN

We have fixed an issue in `%prep` section processing. For instance, if the `%patches` macro appeared there, it would have been converted to `%patch es`, causing failure when executing `%prep` later.

RELEASE NOTES END

### DIFF
--- a/specfile/prep.py
+++ b/specfile/prep.py
@@ -357,6 +357,10 @@ class Prep(collections.abc.Container):
                     m.group("d"),
                     m.group("o"),
                 )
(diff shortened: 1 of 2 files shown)


majamassarini (approved):

lachmanfrantisek (approved):
Thanks!

PR #361, opened 2024-04-03
Handle trailing newline in macro definitions

### LINKED ISSUE #360: MacroDefinition ignore body trailing newline
### What happened? What is the problem?

Macro body may have a trailing newline like [this](https://src.fedoraproject.org/rpms/qemu/blob/fbaca254af85d42563590c443ea2f18c34e8d6f8/f/qemu.spec#_302). In such case specfile lib may fail in calculation of macro positions and can not parse section tags:
```
  File "/usr/lib/python3.6/site-packages/specfile/context_management.py", line 133, in __call__
    self.values[key] = next(self.generators[key])
  File "/usr/lib/python3.6/site-packages/specfile/specfile.py", line 272, in tags
    tags = Tags.parse(section, context=self)
  File "/usr/lib/python3.6/site-packages/specfile/tags.py", line 497, in parse
    lines = process_conditions(list(section), macro_definitions, context)
  File "/usr/lib/python3.6/site-packages/specfile/conditions.py", line 119, in process_conditions
    result.append((line, branches[-2]))
IndexError: list index out of range
```

### What did you expect to happen?

No IndexError

### Example URL(s)

_No response_

### Steps to reproduce

```bash
1. get https://src.fedoraproject.org/rpms/qemu/raw/fbaca254af85d42563590c443ea2f18c34e8d6f8/f/qemu.spec
2. try to get tags of 'package' section


from specfile import Specfile
with Specfile('qemu.spec', sources_path) as specfile:
    specfile.tags('package')
```
```


### Workaround

- [ ] There is an existing workaround that can be used until this issue is fixed.

### Participation

- [X] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

nikitych:
Fixes #360.

### DIFF
--- a/specfile/conditions.py
+++ b/specfile/conditions.py
@@ -82,7 +82,7 @@ def process_conditions(
     if macro_definitions:
         for md in macro_definitions:
             position = md.get_position(macro_definitions)
-            excluded_lines.append(range(position, position + len(md.body.splitlines())))
+            excluded_lines.append(range(position, position + len(md.body.split("\n"))))
     condition_regex = re.compile(
         r"""
         ^
--- a/specfile/macro_definitions.py
+++ b/specfile/macro_definitions.py
@@ -156,11 +156,7 @@ class MacroDefinition:
             elif self.comment_out_style is CommentOutStyle.HASH:
                 sc = "#"
         macro = "global" if self.is_global else "define"
-        body = self.body.splitlines()
-        if body:
-            body[-1] += ws[3]
-        else:
-            body = [ws[3]]
+        body = (self.body + ws[3]).split("\n")
         result.append(f"{ws[0]}{dnl}{pre}{sc}{macro}{ws[1]}{self.name}{ws[2]}{body[0]}")
         result.extend(body[1:])
         return result
@@ -383,7 +379,7 @@ class MacroDefinitions(UserList[MacroDefinition]):
                         line, _ = pop(lines)
                         body += "\n" + line
                         bc, pc = count_brackets(body)
-                tokens = re.split(r"(\s+)$", body, maxsplit=1)
+                tokens = re.split(r"([^\S\n]+)$", body, maxsplit=1)
                 if len(tokens) == 1:
                     body = tokens[0]
                 else:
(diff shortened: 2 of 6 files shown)


nforro (approved):
Thank you for the fix!

However, I believe this is also a parsing issue. The ending newline character should be part of the body, not included in `_whitespace`. To make that right, the regex here should be replaced with `([^\S\n]+)$` (not to match `\n` characters):

https://github.com/packit/specfile/blob/a14e6887852b6e831e3a0c4f9c4230768f2afc4d/specfile/macro_definitions.py#L386

If you are fine with it, I will merge the PR as it is and fix the parsing myself afterwards.

nikitych:
I'll rewrite PR to add `\n` to the body.

There are also a couple of technically errors of missing `\n` [here](https://github.com/packit/specfile/blob/a14e6887852b6e831e3a0c4f9c4230768f2afc4d/specfile/conditions.py#L85) and [here](https://github.com/packit/specfile/blob/a14e6887852b6e831e3a0c4f9c4230768f2afc4d/specfile/sections.py#L271). I spent some time but could not figure out how to exploit the non-excluded empty string. So I could fix them as well, but without adding tests.

nforro:
> I'll rewrite PR to add `\n` to the body.

Thanks! Would you also add that case to [`test_parse`](https://github.com/packit/specfile/blob/a14e6887852b6e831e3a0c4f9c4230768f2afc4d/tests/unit/test_macro_definitions.py#L87)?

> There are also a couple of technically errors of missing `\n` [here](https://github.com/packit/specfile/blob/a14e6887852b6e831e3a0c4f9c4230768f2afc4d/specfile/conditions.py#L85) and [here](https://github.com/packit/specfile/blob/a14e6887852b6e831e3a0c4f9c4230768f2afc4d/specfile/sections.py#L271). I spent some time but could not figure out how to exploit the non-excluded empty string. So I could fix them as well, but without adding tests.

Good point, but I don't think it matters here, empty lines don't have to be excluded because they can't possibly contain a start of a section. But feel free to replace `splitlines()` with `split("\n")` (is that the fix you had in mind?).

nforro (approved):
Thanks!

PR #374, opened 2024-05-09
Process seemingly commented-out macro definitions correctly

nforro:
It may seem that `#%global prerel rc1` is a commented-out macro definition, however as the `%` is not escaped it is still being evaluated and the `prerel` macro becomes defined.

Macro definitions like this are now parsed correctly, having the `commented_out` attribute set to `False`.


lbarcziova (approved):
:+1:

PR #376, opened 2024-05-10
Match only actual source and patch tags

nforro:
Adjust the conditions so that only tags named "Source" or "Patch" optionally followed by a number are matched and for example the "SourceLicense" tag is not.


lbarcziova (approved):

nforro:
There is dnf5 in rawhide again and `dnf config-manager` doesn't work as it used to, hence the test failure(s). I'm working on a fix, but since it's unrelated to this PR I'm going to merge it.

PR #379, opened 2024-05-22
Make (N)EVR(A) objects comparable

nforro:
~~Related to https://github.com/packit/packit-service/issues/2378.~~ In the end it will probably not be needed for the sidetags related stuff, but it can be useful nevertheless.


lbarcziova (approved):
nice!

PR #380, opened 2024-06-04
Improve compatibility with rpm 4.20alpha

nforro:
rpm 4.20alpha (now in rawhide) obsoletes `%patchN` macros and spec files containing them are unparseable. Stop using such macros in integration tests and add support for the `%patch N` variant.

RELEASE NOTES BEGIN

Improved compatibility with RPM 4.20 (alpha version is currently in Fedora Rawhide).

RELEASE NOTES END

### DIFF
--- a/specfile/prep.py
+++ b/specfile/prep.py
@@ -112,19 +112,26 @@ class PatchMacro(PrepMacro):
     @property
     def number(self) -> int:
         """Number of the %patch macro."""
-        if self.options.P is not None:
-            return int(self.options.P)
         tokens = re.split(r"(\d+)", self.name, maxsplit=1)
         if len(tokens) > 1:
             return int(tokens[1])
+        if self.options.P is not None:
+            return int(self.options.P)
+        if self.options.positional:
+            return int(self.options.positional[0])
         return 0
 
     @number.setter
     def number(self, value: int) -> None:
-        if self.options.P is not None:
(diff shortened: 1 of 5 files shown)


specfile/prep.py
```diff
@@ -112,19 +112,26 @@ class PatchMacro(PrepMacro):
     @property
     def number(self) -> int:
         """Number of the %patch macro."""
-        if self.options.P is not None:
-            return int(self.options.P)
         tokens = re.split(r"(\d+)", self.name, maxsplit=1)
         if len(tokens) > 1:
             return int(tokens[1])
```
mfocko, line 117: Can the `-P` option be in other than the first place? Cause I have a feeling that for this:
```
%patch -p2 -P28 -b .patch28
```

It would yield the `2` instead of the `28` :eyes:

nforro: It can, but I'm confused - these two lines are for when the number is part of macro name, e.g. `%patch4`.
It is checked first because that's the priority of expansion: number in the name (with RPM < 4.20), number as the `-P` option and finally number as a positional argument.

nforro: > It would yield the `2` instead of the `28` 👀

It wouldn't, `-p` is not `-P` and order doesn't matter.

mfocko: Oh, that's on the `.name`, not the whole line… got it

mfocko (approved):

lbarcziova (approved):

PR #385, opened 2024-06-17
Test EPEL 8 builds on RHEL 8 compose

nforro:
copr-rpmbuild still builds against EPEL 8 and depends on specfile so we can't drop the support.
Use RHEL 8 compose to test EPEL 8 builds.


lbarcziova (approved):
:+1:

PR #388, opened 2024-06-25
Avoid updating entity value if the entity is referenced from the new value

nforro:
Consider the folowing excerpt from a spec file:

```rpm-spec
%global rel 1%{?dist}

Release: %{rel}
```

Previously, calling `Specfile.update_tag("Release", "%rel")` resulted in the macro definition being updated to `%global rel %rel`, making the spec file invalid. After this change, the macro definition stays unchanged and the tag is updated to `Release: %rel`.


lbarcziova (approved):
lgtm

PR #390, opened 2024-06-26
drop python<3=3.7 support

kloczek:
According to https://endoflife.date/python python 3.7 has been EOSed 27 Jun 2023.
Filter all code over `pyupgracde --py38-plus`.

TODO:

- [ ] Write new tests or update the old ones to cover new functionality.
- [ ] Update doc-strings where appropriate.
- [ ] Update or write new documentation in `packit/packit.dev`.
- [ ] ‹fill in›

Fixes

Related to

Merge before/after

RELEASE NOTES BEGIN

Packit now supports automatic ordering of ☕ after all checks pass.

RELEASE NOTES END

### DIFF
--- a/specfile/changelog.py
+++ b/specfile/changelog.py
@@ -400,15 +400,13 @@ def guess_packager() -> str:
     if shutil.which("git"):
         email = subprocess.run(
             ["git", "config", "user.email"],
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-            universal_newlines=True,
+            capture_output=True,
+            text=True,
         ).stdout.strip()
         fullname = subprocess.run(
             ["git", "config", "user.name"],
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-            universal_newlines=True,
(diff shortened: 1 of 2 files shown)


nforro:
Upstream may not support Python 3.7 and older versions anymore, but Red Hat supports Python 3.6 in RHEL 8 and this library is still being built for EPEL 8.

kloczek:
I don't think that it will be possible to hold that support integrated in master code until 2029 (when EPEL 8 will be EOLed).

IMO better would be just assume that it it will be any critical issue in last used by EPEL 8 version always it can be used last release ta to create branch -> apply necessary fix -> release fixed legacy version.
Many fast moving python modules already is using that strategy and it allow to keep on master support for only >=3.9 or even >=3.11.

Man/hours overhead on some additional commits in possible legacy branches is usually negligible. At the same time it allows keep very clean/smaller code on master.

nforro:
Good point. But in that case I'd rather drop 3.7 and 3.8 as well, supporting 3.9+ would allow for code simplification in some places.

PR #399, opened 2024-07-17
Implement `Specfile.bump_release()` method

### LINKED ISSUE #364: add rpmdev-bumpspec logic
### Description

Right now, the user has to implement release bumping, which can be quite complicated. It would be nice if we could copy logic from rpmdev-bumpspec. That would honor autorelease if used. And simply bump up something by one if needed.

### Benefit

User does not need to study what can be in release and what should be bumped. Logic in one place.The user does not need to study what can be in the release and what should be bumped.

### Importance

I have a workaround right now.

### Workaround

- [X] There is an existing workaround that can be used until this feature is implemented.

### Participation

- [ ] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

nforro:
Fixes #364.

RELEASE NOTES BEGIN

It is now possible to bump a release in a manner similar to `rpmdev-bumpspec` using `Specfile.bump_release()` method.

RELEASE NOTES END

### DIFF
--- a/specfile/specfile.py
+++ b/specfile/specfile.py
@@ -989,3 +989,50 @@ class Specfile:
             return version[:base_end] + "~" + version[suffix_start:]
 
         self.update_tag("Version", handle_prerelease(version))
+
+    @staticmethod
+    def _bump_release_string(release_string: str) -> str:
+        """
+        Bumps release string. Follows the logic of `rpmdev-bumpspec`.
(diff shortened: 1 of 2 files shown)


xsuchy (approved):
+1
I tried it on some spec files I found tricky in past and no problem so far.
Thank you.

PR #406, opened 2024-07-29
Enable use of sidetags for downstream automation

nforro:


.packit.yaml
```diff
@@ -187,17 +187,24 @@ jobs:
...
 
   - job: bodhi_update
     trigger: commit
-    packages: [specfile]
+    packages: [specfile-epel8]
     packit_instances: ["stg"]
     allowed_builders: ["packit-stg", "packit"]
     dist_git_branches:
-      - fedora-branched # rawhide updates are created automatically
-      - epel-9
```
majamassarini, line 202: This is just temporary, right? Because we are missing the bodhi updates for the side tags?

majamassarini (approved):

.packit.yaml
nforro: No, this is Bodhi update for `specfile-epel8` package, neither ogr nor packit support EPEL8 anymore.

majamassarini: now I see the bodhi job in the packit package...

nforro: `bodhi_update` job for ogr, specfile and packit is configured only in packit.

PR #409, opened 2024-09-04
Explicitly invalidate the global parse hash in `SpecParser` constructor

nforro:
Python can reuse id of objects after they are garbage-collected, however the global parse hash contains an id of `SpecParser` instance in order to force parsing on context switches, and when a different instance has the same id, it has no way to detect that.

Explicitly invalidate the global parse hash when a `SpecParser` instance is created to prevent this issue.

Fixes https://github.com/packit/packit-service/issues/2461.


lbarcziova (approved):
interesting, good catch!

PR #423, opened 2024-11-12
Allow building specfile in CentOS Stream without EPEL

majamassarini:
CentOS Stream does not have EPEL packages in CBS Koji.
Package typing-extensions is not available.

### DIFF
--- a/specfile/types.py
+++ b/specfile/types.py
@@ -3,10 +3,12 @@
 
 import abc
 
-from typing_extensions import Protocol
+try:
+    from typing import SupportsIndex
+except ImportError:
+    # define our own SupportsIndex type for older version of typing (Python 3.7 and older)
+    from typing_extensions import Protocol
 
-
-# define our own SupportsIndex type for older version of typing_extensions (EL 8)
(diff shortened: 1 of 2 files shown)


nforro (approved):
:+1: 
Just a nitpick about the commit message. I would change it to `Allow building specfile in CentOS Stream without EPEL`. We regularly build specfile in CentOS Stream, just with EPEL repos enabled.

specfile/types.py
```diff
@@ -3,10 +3,12 @@
...
-
-
-# define our own SupportsIndex type for older version of typing_extensions (EL 8)
-class SupportsIndex(Protocol, metaclass=abc.ABCMeta):
-    @abc.abstractmethod
-    def __index__(self) -> int: ...
+try:
+    from typing import SupportsIndex
+except ImportError:
+    # define our own SupportsIndex type for older version of typing_extensions (EL 8)
```
nforro, line 9: Actually, one more nitpick:

```suggestion
    # define our own SupportsIndex type for older version of typing (Python 3.7 and older)
```

PR #439, opened 2024-12-11
Build in Copr for aarch64

nforro:
`fedora-stable` should be sufficient for container image builds. but perhaps it would make more sense to use `fedora-all-aarch64` in jobs where we already build for `fedora-all`? WDYT?


majamassarini (approved):

nforro:
Ok, I've changed the chroots to match, and let's enable ppc64le when it's needed, to avoid potential long queues in Copr.

Issue #404, opened 2024-07-28
AttributeError: 'Sections' object has no attribute 'changelog'


xsuchy:
### Description

When Specfile does not have changelog section and you try to read it you get:

```
Traceback (most recent call last):
  File "/home/msuchy/projects/license-validate/./print-spec-changelog.py", line 10, in <module>
    print(sections.changelog)
          ^^^^^^^^^^^^^^^^^^
  File "/home/msuchy/.local/lib/python3.12/site-packages/specfile/sections.py", line 180, in __getattr__
    return super().__getattribute__(id)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Sections' object has no attribute 'changelog'
```

The script is 
```
#!/usr/bin/python3
from specfile import Specfile
import sys

# this is dump, but do the work
filename = sys.argv[1]
specfile = Specfile(filename, force_parse=True)

with specfile.sections() as sections:
    print(sections.changelog)
```

While the TB is technically correct, I would argue that the getter should return empty section to make the user experience better.

### Workaround

- [X] There is an existing workaround that can be used until this feature is implemented.

### Participation

- [ ] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

nforro:
I think the library should be able to express the difference between a spec file without a `%changelog` and a spec file with an empty `%changelog`. While it could be possible to "decouple" `"changelog" in sections` from `sections.changelog`, I'm not sure it's the best idea. Also, if `sections.changelog` returned an empty section, it would technically be creating a new section in the spec file, which should be a conscious action.

nforro:
How about something like this, to make the creation explicit:

```python
with spec.sections() as sections:
    changelog = sections.get_or_create("changelog")
    changelog[:] = ["%autochangelog"]
```

which would be equivalent to this:

```python
with spec.sections() as sections:
    try:
        sections.changelog[:] = ["%autochangelog"]
    except AttributeError:
        sections.append(Section("changelog", data=["%autochangelog"]))
```

xsuchy:
I like it.

PR #441, opened 2024-12-12
Implement `Sections.get_or_create()`

nforro:
Fixes https://github.com/packit/specfile/issues/404.

RELEASE NOTES BEGIN

There is a new convenience method `Sections.get_or_create()` that allows you to manipulate a section without checking if it exists first. If a section doesn't exist, it will be appended to the end. For example, this will work properly even on spec files without `%changelog`:

```
with spec.sections() as sections:
    changelog = sections.get_or_create("changelog")
    changelog[:] = ["%autochangelog"]
```

RELEASE NOTES END

### DIFF
--- a/specfile/sections.py
+++ b/specfile/sections.py
@@ -214,6 +214,38 @@ class Sections(UserList[Section]):
                 return i
         raise ValueError
 
+    def get_or_create(self, id: str) -> Section:
+        """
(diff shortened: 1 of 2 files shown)


lachmanfrantisek (approved):
Seems handy. Thanks for the example in the release notes.

PR #444, opened 2024-12-18
Add support for openSUSE style detached changelogs

dcermak:
This PR adds some hacky support for detached/openSUSE style changes files that are inserted into the %changelog section by the open build service.

The tests are not yet all passing, as there is a discrepancy between `str(DetachedChangelog)` and respective section in `str(spec)`. I'm not sure if this is caused by my (probably rather odd) approach here or whether this is even required?

TODO:

- [ ] Update or write new documentation in `packit/packit.dev`.

Fixes

Related to https://github.com/packit/packit/issues/2399

Merge before/after

RELEASE NOTES BEGIN

Add support for detached (open)SUSE style changelogs

RELEASE NOTES END

### DIFF
--- a/specfile/changelog.py
+++ b/specfile/changelog.py
@@ -9,6 +9,7 @@ import pwd
 import re
 import shutil
 import subprocess
+from enum import Enum, auto, unique
 from typing import List, Optional, Union, overload
 
 from specfile.exceptions import SpecfileException
@@ -33,6 +34,24 @@ MONTHS = (
     "Nov",
     "Dec",
 )
+
+_OPENSUSE_CHANGELOG_SEPARATOR = 67 * "-"
+
+
+@unique
+class ChangelogStyle(Enum):
+    """Style of changelog entries"""
+
+    #: standard changelog entries parseable by RPM (used in Fedora, RHEL, etc.):
+    #: * $DATE $AUTHOR <$EMAIL> - $EVR
+    #: $ENTRY
+    standard = auto()
+
+    #: openSUSE/SUSE style detached changelog:
+    #: -------------------------------------------------------------------
+    #: $DATE - $AUTHOR <$EMAIL>
+    #: $ENTRY
+    openSUSE = auto()
 
 
 class ChangelogEntry:
@@ -168,6 +187,13 @@ class ChangelogEntry:
             return ""
         return m.group("wsp") + (m.group("zp") or "")
 
+    @property
+    def style(self) -> ChangelogStyle:
+        """Style of this changelog entry (standard vs openSUSE)."""
+        if self.header.startswith(_OPENSUSE_CHANGELOG_SEPARATOR):
+            return ChangelogStyle.openSUSE
+        return ChangelogStyle.standard
+
     @classmethod
     def assemble(
         cls,
@@ -177,6 +203,7 @@ class ChangelogEntry:
         evr: Optional[str] = None,
         day_of_month_padding: str = "0",
         append_newline: bool = True,
+        style: ChangelogStyle = ChangelogStyle.standard,
     ) -> "ChangelogEntry":
         """
         Assembles a changelog entry.
@@ -184,30 +211,55 @@ class ChangelogEntry:
         Args:
             timestamp: Timestamp of the entry.
                 Supply `datetime` rather than `date` for extended format.
+                openSUSE-style changelog entries mandate extended format, so if a `date`
+                is supplied, the timestamp will be set to noon of that day.
             author: Author of the entry.
             content: List of lines forming the content of the entry.
             evr: EVR (epoch, version, release) of the entry.
+                Ignored if `style` is `ChangelogStyle.openSUSE`.
             day_of_month_padding: Padding to apply to day of month in the timestamp.
             append_newline: Whether the entry should be followed by an empty line.
+            style: Which style of changelog should be created.
 
         Returns:
             New instance of `ChangelogEntry` class.
         """
         weekday = WEEKDAYS[timestamp.weekday()]
         month = MONTHS[timestamp.month - 1]
-        header = f"* {weekday} {month}"
+
+        if style == ChangelogStyle.standard:
+            header = "* "
+        else:
+            header = _OPENSUSE_CHANGELOG_SEPARATOR + "\n"
+        header += f"{weekday} {month}"
+
         if day_of_month_padding.endswith("0"):
             header += f" {day_of_month_padding[:-1]}{timestamp.day:02}"
         else:
             header += f" {day_of_month_padding}{timestamp.day}"
+
+        # convert to extended format for openSUSE style changelogs
+        if style == ChangelogStyle.openSUSE and not isinstance(
+            timestamp, datetime.datetime
+        ):
+            timestamp = datetime.datetime(
+                year=timestamp.year, month=timestamp.month, day=timestamp.day, hour=12
+            )
+
         if isinstance(timestamp, datetime.datetime):
             # extended format
             if not timestamp.tzinfo:
                 timestamp = timestamp.replace(tzinfo=datetime.timezone.utc)
             header += f" {timestamp:%H:%M:%S %Z}"
-        header += f" {timestamp:%Y} {author}"
-        if evr is not None:
+        header += f" {timestamp:%Y} "
+
+        if style == ChangelogStyle.openSUSE:
+            header += "- "
+        header += author
+
+        if evr is not None and style == ChangelogStyle.standard:
             header += f" - {evr}"
+
         return cls(header, content, [""] if append_newline else None)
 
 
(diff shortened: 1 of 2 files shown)


nforro:
I appreciate the effort, but I don't think this should be part of `Specfile` - it represents a single `*.spec` file, it shouldn't read or modify any other files. Also, I don't understand why there is a need for OpenSUSE-specific classes. If OpenSUSE changelog entries are parseable by RPM, `ChangelogEntry` should be able to parse/construct them without issues.

dcermak:
> I appreciate the effort, but I don't think this should be part of `Specfile` - it represents a single `*.spec` file, it shouldn't read or modify any other files. 

Yes, I understand that, the problem is that in openSUSE land it has been the convention for ever and ever to have the changelog in a secondary file and not be it part of the `*.spec` file. That in itself wouldn't be a problem, but since we would like to extend packit to also handle openSUSE, we need to be able to append to the changelog. Ideally, the API should be the same for all distros as much as possible, so that we don't have to have multiple code paths.

> Also, I don't understand why there is a need for OpenSUSE-specific classes. If OpenSUSE changelog entries are parseable by RPM, `ChangelogEntry` should be able to parse/construct them without issues.

Yes, the openSUSE changes files follow a different format. A typical changes file entry looks like this:
```
-------------------------------------------------------------------
Wed Dec 18 10:49:41 UTC 2024 - Dan Čermák <dcermak@suse.com>

- Enable tests
```
In contrast to the typical rpm changelog, it always contains a time in UTC, but lacks the EVR (for practical reasons as the build service auto-rebuilds bumping the release).

The build service or obs-build or rpmbuild add the changes file into the built rpm, but that happens on a binary rpm level. This doesn't really help in the context of packit.

nforro:
> Yes, I understand that, the problem is that in openSUSE land it has been the convention for ever and ever to have the changelog in a secondary file and not be it part of the `*.spec` file. That in itself wouldn't be a problem, but since we would like to extend packit to also handle openSUSE, we need to be able to append to the changelog. Ideally, the API should be the same for all distros as much as possible, so that we don't have to have multiple code paths.
> 

I would argue it should be part of packit codebase, there is literally a class called [ChangelogHelper](https://github.com/packit/packit/blob/e5b3d8c94ec79aa9cc5e1fa8dd7d0aabc1dcfbf4/packit/utils/changelog_helper.py#L17).

> Yes, the openSUSE changes files follow a different format. A typical changes file entry looks like this:
> 
> ```
> -------------------------------------------------------------------
> Wed Dec 18 10:49:41 UTC 2024 - Dan Čermák <dcermak@suse.com>
> 
> - Enable tests
> ```

I see.

> In contrast to the typical rpm changelog, it always contains a time in UTC, but lacks the EVR (for practical reasons as the build service auto-rebuilds bumping the release).

That shouldn't be a problem, even some Fedora/RHEL spec files don't have EVR in changelog entries. AFAICT the only difference is the line of dashes instead of an asterisk denoting the start of an entry. IMHO it would make sense to add support for that into `ChangelogEntry` and `Changelog.parse()` and keep the rest as it is.

nforro:
To be a bit more specific, `Changelog.parse()` would be able to parse both formats and corresponding `ChangelogEntry.header` attributes would start either with `* ` or with `-------------------------------------------------------------------\n`, and `ChangelogEntry.assemble()` would gain a new argument to choose the correct "starting sequence".

dcermak:
Ok, I'll see how far I get then.

How would you recommend to handle the parsing/creation of `*.changes` files? Because if we read a specfile from OBS using the specfile library, it'll have an empty `%changelog`. And anything that packit creates, will have the changelog in the spec and thereby get auto-rejected for inclusion in the openSUSE distributions.

nforro:
> How would you recommend to handle the parsing/creation of `*.changes` files? Because if we read a specfile from OBS using the specfile library, it'll have an empty `%changelog`. And anything that packit creates, will have the changelog in the spec and thereby get auto-rejected for inclusion in the openSUSE distributions.

Well, packit should do something similar to what `Specfile.add_changelog_entry()` does, except it won't be operating on `*spec` but on `*.changes` (although there won't be much left from the original implementation apart from `ChangelogEntry.assemble()`).
https://github.com/packit/specfile/blob/c0a98a85d9e43f59447248bb7f5258c2406ce9f1/specfile/specfile.py#L479-L551

I mean, we could also add an argument to `Specfile.add_changelog_entry()` specifying path to a changelog file and then the method would write the entry to that file instead to the spec file itself, but I'm still not convinced such code belongs here.

nforro:
Fedora packages can use changelog files too BTW ([example](https://src.fedoraproject.org/rpms/man-pages/blob/rawhide/f/changelog)), but they exist for preserving changelog history and are not meant to be updated.

dcermak:
@nforro I have reworked the PR, what do you think about this approach? I hope it's as noninvasive as possible

specfile/changelog.py
```diff
@@ -34,6 +35,26 @@
     "Dec",
 )
 
+_OPENSUSE_CHANGELOG_SEPARATOR = (
+    "-------------------------------------------------------------------"
+)
```
nforro, line 40:
```suggestion
_OPENSUSE_CHANGELOG_SEPARATOR = 67 * "-"
```

specfile/changelog.py
```diff
@@ -177,25 +198,42 @@ def assemble(
...
+        style: ChangelogStyle = ChangelogStyle.standard,
     ) -> "ChangelogEntry":
         """
         Assembles a changelog entry.
 
         Args:
             timestamp: Timestamp of the entry.
                 Supply `datetime` rather than `date` for extended format.
+                The hour is set to 12 for openSUSE style changelog entries, if a
+                date is passed instead of a datetime
```
nforro, line 210:
```suggestion
                openSUSE-style changelog entries mandate extended format, so if a `date`
                is supplied, the timestamp will be set to noon of that day.
```

specfile/changelog.py
```diff
@@ -177,25 +198,42 @@ def assemble(
...
             author: Author of the entry.
             content: List of lines forming the content of the entry.
             evr: EVR (epoch, version, release) of the entry.
```
nforro, line 213:
```suggestion
            evr: EVR (epoch, version, release) of the entry.
                Ignored if `style` is `ChangelogStyle.openSUSE`.
```

specfile/changelog.py
```diff
@@ -177,25 +198,42 @@ def assemble(
...
             day_of_month_padding: Padding to apply to day of month in the timestamp.
             append_newline: Whether the entry should be followed by an empty line.
+            style: Which style of changelog should be created
```
nforro, line 216:
```suggestion
            style: Which style of changelog should be created.
```

specfile/changelog.py
```diff
@@ -177,25 +198,42 @@ def assemble(
...
         Returns:
             New instance of `ChangelogEntry` class.
         """
+        if style == ChangelogStyle.openSUSE and not isinstance(
+            timestamp, datetime.datetime
+        ):
+            timestamp = datetime.datetime(
+                year=timestamp.year, month=timestamp.month, day=timestamp.day, hour=12
+            )
+
```
nforro, line 227: I would move this down a bit, where `timestamp` is handled, it doesn't make much sense to me to have it here.

specfile/changelog.py
```diff
@@ -350,13 +394,24 @@ def extract_following_lines(content: List[str]) -> List[str]:
...
         content: List[str] = []
-        for line in section:
-            if line.startswith("*"):
+
+        for i, line in enumerate(section):
+            if line == _OPENSUSE_CHANGELOG_SEPARATOR:
+                continue
+
+            if line.startswith("*") or (
+                i >= 1 and section[i - 1] == _OPENSUSE_CHANGELOG_SEPARATOR
```
nforro, line 403: The same condition is used twice, it would make sense to store the result into a variable.

nforro (approved):
Thanks! LGTM.
Nice to have would be a property of `ChangelogEntry` that would return `ChangelogStyle`.

nforro:
Just wondering, is rpm in openSUSE patched to be able to parse the changelogs (i.e. does something like `cat *.spec *.changes | rpmspec --parse /dev/stdin` work)?

specfile/changelog.py
```diff
@@ -34,6 +35,24 @@
...
 )
 
+_OPENSUSE_CHANGELOG_SEPARATOR = 67 * "-"
+
+
+@unique
+class ChangelogStyle(Enum):
+    """Style of changelog entries"""
+
+    #: "standard" changelog entries as used by Fedora, RHEL, etc.:
```
nforro, line 45: I would mention that the standard style is what RPM is able to parse (unless the openSUSE style is also parseable):

```suggestion
    #: standard changelog entries parseable by RPM (used in Fedora, RHEL, etc.):
```

nforro (approved):

nforro:
> ```
> # cat container-snap.changes | rpmspec --parse /dev/stdin
> error: line 1: Unknown tag: -------------------------------------------------------------------
> ```

I meant when you (con)cat a specfile (without changelog) with the corresponding changes file (and add `%changelog` inbetween if it's missing entirely). I wouldn't expect a sole `.changes` file to be parseable.

dcermak:
> > ```
> > # cat container-snap.changes | rpmspec --parse /dev/stdin
> > error: line 1: Unknown tag: -------------------------------------------------------------------
> > ```
> 
> I meant when you (con)cat a specfile (without changelog) with the corresponding changes file (and add `%changelog` inbetween if it's missing entirely). I wouldn't expect a sole `.changes` file to be parseable.

Oh, dang. That's what you meant. So `rpmspec` appears to parse it (it returns 0), but it throws a warning:
```ShellSession
# cat /src/container-snap.spec /src/container-snap.changes | rpmspec --parse /dev/stdin > /dev/null
error: %changelog entries must start with *
```

Afaik there is some logic in the build service that masages the changes files back into the binary rpms

nforro:
> Afaik there is some logic in the build service that masages the changes files back into the binary rpms

Right, that's what I thought. Thanks for the investigation and for the PR :slightly_smiling_face:

PR #450, opened 2025-01-28
Remove the walrus operator for Python 3.6 compatibility

nforro:
Related to https://github.com/packit/specfile/pull/444.

We still support EL8 and thus Python 3.6. CI didn't catch this because it didn't even run on RHEL8 (`dcermak can't run tests (and builds) internally` - is there a way for a repo owner to overcome this?).

### DIFF
--- a/tests/unit/test_changelog.py
+++ b/tests/unit/test_changelog.py
@@ -257,47 +257,35 @@ def test_suse_style_changelog_parse():
             "changelog",
             data=[
                 "-------------------------------------------------------------------",
-                (
-                    hdr1 := "Tue Dec 17 14:21:37 UTC 2024 - "
-                    + (dc := "Dan Čermák <dan.cermak@cgc-instruments.com>")
-                ),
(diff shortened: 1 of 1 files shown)


lbarcziova (approved):
> dcermak can't run tests (and builds) internally - is there a way for a repo owner to overcome this?

I would say just by manual comment triggering

mfocko (approved):

majamassarini (approved):
it is just a test, so it is fine as it is.
But I am wondering why don't move the variables declarations at the beginning of the test, in an old style way, and use them later (instead of repeating the strings?)

nforro:
> But I am wondering why don't move the variables declarations at the beginning of the test, in an old style way, and use them later (instead of repeating the strings?)

You mean outside of test functions? I think that would make the readability worse.

lbarcziova:
> Right, I actually did that on https://github.com/packit/specfile/pull/445 but the test jobs kept erroring so I didn't even try on the other PR. But the incompatibility would have been caught already during build, so I should have done it anyway.

we could also add the checks at least for builds to the required ones for merging

PR #457, opened 2025-03-11
context_management: add a type stub override to fix typing

gotmax23:
Type checkers (mypy, pyright, et al.) don't understand the ContextManager descriptor class.
With a stub file, we can tell the type checker to treat ContextManager as a simple decorator function which is something that it understands.

``` python
from typing import reveal_type

import specfile

s = specfile.Specfile("./fedora/python-specfile.spec")

# Before: types.MethodType
# After: (allow_duplicates: bool = False, default_to_implicit_numbering: bool = False, default_source_number_digits: int = 1) -> GeneratorContextManager[Sources]
reveal_type(s.sources)
# Before: Any
# After: GeneratorContextManager[Sources]
reveal_type(s.sources())
# Before: Any
# After: Sources
reveal_type(s.sources().__enter__())
reveal_type(s.sources().content)
```

RELEASE NOTES BEGIN

context_management: add a type stub override to fix typing. Type checkers like mypy and pyright can now correctly determine the types for `.sources()`, `.sections()`, and the other `Specfile` methods that return context managers.

RELEASE NOTES END

### DIFF
--- a/specfile/context_management.pyi
+++ b/specfile/context_management.pyi
@@ -1,0 +1,25 @@
+# Copyright Contributors to the Packit project.
+# SPDX-License-Identifier: MIT
+
+import contextlib
+from collections.abc import Iterator
+from typing import Callable, Generator, ParamSpec, TypeVar
+
+_T_co = TypeVar("_T_co", covariant=True)
+_P = ParamSpec("_P")
+
+@contextlib.contextmanager
+def capture_stderr() -> Generator[list[bytes], None, None]: ...
+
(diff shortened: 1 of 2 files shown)


specfile/context_management.pyi
```diff
@@ -0,0 +1,22 @@
+import contextlib
```
nforro, line 1:
```suggestion
# Copyright Contributors to the Packit project.
# SPDX-License-Identifier: MIT

import contextlib
```

nforro (approved):
This is great, thanks!

Could you just add the copyright header to the new file?

nforro:
regate

PR #458, opened 2025-03-12
ft: Instantiate Specfile from strings or file-like objects

### LINKED ISSUE #206: Create Specfile from a virtual file
Currently, it is possible to create an object of `Specfile` like this:

```python
# using an absolute path
specfile = Specfile('/tmp/test.spec')

# using a relative path and a different sourcedir
specfile = Specfile('test.spec', sourcedir='/tmp/sources')
```

It would be nice to be able to create it also like this

```python
with open('/tmp/test.spec', 'r') as fp:
    specfile = Specfile(fp)
```

This alone is not so interesting, but it allows for creating `Specfile` from a virtual file

```python
from io import StringIO
fp = StringIO()
fp.write("Name: foo\nVersion: 1.0\nRelease: 1\n...")
specfile = Specfile(fp)
```

Of course, I am interested in the feature, not the exact syntax. My suggestion will probably have an issue with the `.save()` method so if it is done like

```python
specfile = Specfile(fp=fp)

# or

specfile = Specfile.from_fp(fp)
```
It is all fine with me :-)

### LINKED ISSUE #248: Instantiate Specfile with a string
### Description

Currently, it's possible to do this via a file path. In my use case, all I'm doing is fetching the specfile for multiple packages via OSC API and then trying to parse them using Specfile. I'm currently creating a temp file and then passing it on to the package.

Is this something that can be supported or have there been any discussions in the past about this that ruled it out? 

### Benefit

_No response_

### Importance

_No response_

### Workaround

- [X] There is an existing workaround that can be used until this feature is implemented.

### Participation

- [X] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

mynk8:
TODO:

- [x] Update or write new documentation.

Fixes #206 
Fixes #248 

RELEASE NOTES BEGIN

- Added support for creating Specfile instances from file objects and strings.

RELEASE NOTES END

### DIFF
--- a/specfile/specfile.py
+++ b/specfile/specfile.py
@@ -1,13 +1,27 @@
 # Copyright Contributors to the Packit project.
 # SPDX-License-Identifier: MIT
 
+import copy
 import datetime
 import logging
 import re
 import types
 from dataclasses import dataclass
+from io import FileIO, StringIO
 from pathlib import Path
-from typing import Generator, List, Optional, Tuple, Type, Union, cast
+from typing import (
+    IO,
+    Any,
+    Dict,
+    Generator,
+    List,
+    Optional,
+    TextIO,
+    Tuple,
+    Type,
+    Union,
+    cast,
+)
 
 import rpm
 
@@ -31,6 +45,7 @@ from specfile.sourcelist import Sourcelist
 from specfile.sources import Patches, Sources
 from specfile.spec_parser import SpecParser
 from specfile.tags import Tag, Tags
+from specfile.types import EncodingArgs
 from specfile.value_parser import (
     SUBSTITUTION_GROUP_PREFIX,
     ConditionalMacroExpansion,
@@ -50,19 +65,27 @@ class Specfile:
         autosave: Whether to automatically save any changes made.
     """
 
+    ENCODING_ARGS: EncodingArgs = {"encoding": "utf8", "errors": "surrogateescape"}
+
     def __init__(
         self,
-        path: Union[Path, str],
+        path: Optional[Union[Path, str]] = None,
+        content: Optional[str] = None,
+        file: Optional[IO] = None,
         sourcedir: Optional[Union[Path, str]] = None,
         autosave: bool = False,
         macros: Optional[List[Tuple[str, Optional[str]]]] = None,
         force_parse: bool = False,
     ) -> None:
         """
-        Initializes a specfile object.
+        Initializes a specfile object. You can specify either a path to the spec file,
+        its content as a string or a file object representing it. `sourcedir` is optional
+        if `path` or a named `file` is provided and will be set to the parent directory.
 
         Args:
             path: Path to the spec file.
+            content: String containing the spec file content.
+            file: File object representing the spec file.
             sourcedir: Path to sources and patches.
             autosave: Whether to automatically save any changes made.
             macros: List of extra macro definitions.
@@ -71,12 +94,29 @@ class Specfile:
                 Such sources include sources referenced from shell expansions
                 in tag values and sources included using the _%include_ directive.
         """
+        # count mutually exclusive arguments
+        if sum([file is not None, path is not None, content is not None]) > 1:
+            raise ValueError(
+                "Only one of `file`, `path` or `content` should be provided"
+            )
+        if file is not None:
+            self._file = file
+        elif path is not None:
+            self._file = Path(path).open("r+", **self.ENCODING_ARGS)
+        elif content is not None:
+            self._file = StringIO(content)
+        else:
+            raise ValueError("Either `file`, `path` or `content` must be provided")
+        if sourcedir is None:
+            try:
+                sourcedir = Path(self._file.name).parent
+            except AttributeError:
+                raise ValueError(
+                    "`sourcedir` is required when providing `content` or file object without a name"
+                )
         self.autosave = autosave
-        self._path = Path(path)
-        self._lines, self._trailing_newline = self._read_lines(self._path)
-        self._parser = SpecParser(
-            Path(sourcedir or self.path.parent), macros, force_parse
-        )
+        self._lines, self._trailing_newline = self._read_lines(self._file)
+        self._parser = SpecParser(Path(sourcedir), macros, force_parse)
         self._parser.parse(str(self))
         self._dump_debug_info("After initial parsing")
 
@@ -85,7 +125,7 @@ class Specfile:
             return NotImplemented
         return (
             self.autosave == other.autosave
-            and self._path == other._path
+            and self.path == other.path
             and self._lines == other._lines
             and self._parser == other._parser
         )
@@ -111,6 +151,39 @@ class Specfile:
     ) -> None:
         self.save()
 
+    def __deepcopy__(self, memodict: Dict[int, Any]):
+        """
+        Deepcopies the object, handling file-like attributes.
+        """
+        specfile = self.__class__.__new__(self.__class__)
+        memodict[id(self)] = specfile
+
+        for k, v in self.__dict__.items():
+            if k == "_file":
+                continue
+            setattr(specfile, k, copy.deepcopy(v, memodict))
+
+        try:
+            path = Path(cast(FileIO, self._file).name)
+        except AttributeError:
+            # IO doesn't implement getvalue() so tell mypy this is StringIO
+            # (could also be BytesIO)
+            sio = cast(StringIO, self._file)
+            # not a named file, try `getvalue()`
+            specfile._file = type(sio)(sio.getvalue())
+        else:
+            try:
+                # encoding and errors are only available on TextIO objects
+                file = cast(TextIO, self._file)
+                specfile._file = path.open(
+                    mode=file.mode, encoding=file.encoding, errors=file.errors
+                )
+            except AttributeError:
+                # files open in binary mode have no `encoding`/`errors`
+                specfile._file = path.open(self._file.mode)
+
+        return specfile
+
     def _dump_debug_info(self, message) -> None:
         logger.debug(
             f"DBG: {message}:\n"
@@ -119,19 +192,30 @@ class Specfile:
             f"  {self._parser.spec!r} @ 0x{id(self._parser.spec):012x}"
         )
 
-    @staticmethod
-    def _read_lines(path: Path) -> Tuple[List[str], bool]:
-        content = path.read_text(encoding="utf8", errors="surrogateescape")
-        return content.splitlines(), content[-1] == "\n"
+    @classmethod
+    def _read_lines(cls, file: IO) -> Tuple[List[str], bool]:
+        file.seek(0)
+        raw_content = file.read()
+        if isinstance(raw_content, str):
+            content = raw_content
+        else:
+            content = raw_content.decode(**cls.ENCODING_ARGS)
+        return content.splitlines(), content.endswith("\n")
 
     @property
-    def path(self) -> Path:
+    def path(self) -> Optional[Path]:
         """Path to the spec file."""
-        return self._path
+        try:
+            return Path(cast(FileIO, self._file).name)
+        except AttributeError:
+            return None
 
     @path.setter
     def path(self, value: Union[Path, str]) -> None:
-        self._path = Path(value)
+        path = Path(value)
+        if path == self.path:
+            return
+        self._file = path.open("r+", **self.ENCODING_ARGS)
 
     @property
     def sourcedir(self) -> Path:
@@ -179,11 +263,26 @@ class Specfile:
 
     def reload(self) -> None:
         """Reloads the spec file content."""
-        self._lines, self._trailing_newline = self._read_lines(self.path)
+        try:
+            path = Path(cast(FileIO, self._file).name)
+        except AttributeError:
+            pass
+        else:
+            # reopen the path in case the original file has been deleted/replaced
+            self._file.close()
+            self._file = path.open("r+", **self.ENCODING_ARGS)
+        self._lines, self._trailing_newline = self._read_lines(self._file)
 
     def save(self) -> None:
         """Saves the spec file content."""
-        self.path.write_text(str(self), encoding="utf8", errors="surrogateescape")
+        self._file.seek(0)
+        self._file.truncate(0)
+        content = str(self)
+        try:
+            self._file.write(content)
+        except TypeError:
+            self._file.write(content.encode(**self.ENCODING_ARGS))
+        self._file.flush()
 
     def expand(
         self,
--- a/specfile/types.py
+++ b/specfile/types.py
@@ -12,3 +12,12 @@ except ImportError:
     class SupportsIndex(Protocol, metaclass=abc.ABCMeta):  # type: ignore [no-redef]
         @abc.abstractmethod
         def __index__(self) -> int: ...
+
+
+try:
+    from typing import TypedDict
+except ImportError:
+    from typing_extensions import TypedDict
+
+
+EncodingArgs = TypedDict("EncodingArgs", {"encoding": str, "errors": str})
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,10 +1,11 @@
 # Copyright Contributors to the Packit project.
 # SPDX-License-Identifier: MIT
-
+import io
 import shutil
 
 import pytest
 
+from specfile import Specfile
 from tests.constants import (
     SPEC_AUTOPATCH,
     SPEC_AUTOSETUP,
@@ -136,3 +137,46 @@ def spec_conditionalized_version(tmp_path):
     specfile_path = tmp_path / SPECFILE
     shutil.copyfile(SPEC_CONDITIONALIZED_VERSION / SPECFILE, specfile_path)
     return specfile_path
+
+
+@pytest.fixture(
+    params=[
+        "file_path",
+        "text_file",
+        "binary_file",
+        "text_io_stream",
+        "binary_io_stream",
+        "content_string",
+    ]
+)
+def specfile_factory(request):
+    """
+    pytest fixture to create a `Specfile` instance with different input modes.
+
+    Returns:
+        Function that creates a `Specfile` instance.
+    """
+    mode = request.param
+
+    def _create_specfile(input_path, **kwargs):
+        kwargs.setdefault("sourcedir", input_path.parent)
+
+        if mode == "file_path":
+            return Specfile(path=input_path, **kwargs)
+        elif mode == "text_file":
+            f = open(input_path, "r+", **Specfile.ENCODING_ARGS)
+            return Specfile(file=f, **kwargs)
+        elif mode == "binary_file":
+            f = open(input_path, "rb+")
+            return Specfile(file=f, **kwargs)
+        elif mode == "text_io_stream":
+            content = input_path.read_text(**Specfile.ENCODING_ARGS)
+            return Specfile(file=io.StringIO(content), **kwargs)
+        elif mode == "binary_io_stream":
+            content = input_path.read_bytes()
+            return Specfile(file=io.BytesIO(content), **kwargs)
+        elif mode == "content_string":
+            content = input_path.read_text(**Specfile.ENCODING_ARGS)
+            return Specfile(content=content, **kwargs)
+
+    return _create_specfile
--- a/tests/integration/test_specfile.py
+++ b/tests/integration/test_specfile.py
@@ -26,8 +26,8 @@ def test_parse(spec_multiple_sources):
     assert spec.rpm_spec.prep == prep
 
 
-def test_prep_traditional(spec_traditional):
-    spec = Specfile(spec_traditional)
+def test_prep_traditional(specfile_factory, spec_traditional):
+    spec = specfile_factory(spec_traditional)
     with spec.prep() as prep:
         assert AutosetupMacro not in prep.macros
         assert AutopatchMacro not in prep.macros
@@ -50,8 +50,8 @@ def test_prep_traditional(spec_traditional):
         assert sections.prep[1] == "%patch 0 -p2 -b .test2 -E"
 
 
-def test_prep_autosetup(spec_autosetup):
-    spec = Specfile(spec_autosetup)
+def test_prep_autosetup(specfile_factory, spec_autosetup):
+    spec = specfile_factory(spec_autosetup)
     with spec.prep() as prep:
         assert len(prep.macros) == 1
         assert AutosetupMacro in prep.macros
@@ -60,8 +60,8 @@ def test_prep_autosetup(spec_autosetup):
         assert prep.autosetup.options.p == 1
 
 
-def test_prep_autopatch(spec_autopatch):
-    spec = Specfile(spec_autopatch)
+def test_prep_autopatch(specfile_factory, spec_autopatch):
+    spec = specfile_factory(spec_autopatch)
     with spec.prep() as prep:
         assert len(prep.macros) == 4
         assert prep.macros[1].options.M == 2
@@ -75,8 +75,8 @@ def test_prep_autopatch(spec_autopatch):
         assert sections.prep[3] == "%autopatch -p1 0 1 2 3 4 5 6"
 
 
-def test_sources(spec_minimal):
-    spec = Specfile(spec_minimal)
+def test_sources(specfile_factory, spec_minimal):
+    spec = specfile_factory(spec_minimal)
     source = "test.tar.gz"
     with spec.sources() as sources:
         assert not sources
@@ -93,8 +93,8 @@ def test_sources(spec_minimal):
         assert not sources
 
 
-def test_patches(spec_patchlist):
-    spec = Specfile(spec_patchlist)
+def test_patches(specfile_factory, spec_patchlist):
+    spec = specfile_factory(spec_patchlist)
     patch = "test.patch"
     with spec.patches() as patches:
         patches.insert(0, patch)
@@ -214,19 +214,13 @@ def test_patches(spec_patchlist):
     ],
 )
 def test_add_changelog_entry(
-    spec_minimal,
-    entry,
-    author,
-    email,
-    timestamp,
-    evr,
-    result,
+    specfile_factory, spec_minimal, entry, author, email, timestamp, evr, result
 ):
     if author is None:
         flexmock(specfile.specfile).should_receive("guess_packager").and_return(
             "John Doe <john@doe.net>"
         ).once()
-    spec = Specfile(spec_minimal)
+    spec = specfile_factory(spec_minimal)
     spec.add_changelog_entry(entry, author, email, timestamp, evr)
     with spec.sections() as sections:
         assert sections.changelog[: len(result)] == result
@@ -240,8 +234,8 @@ def test_add_changelog_entry(
         ("1.4.6", "0.1rc5"),
     ],
 )
-def test_set_version_and_release(spec_minimal, version, release):
-    spec = Specfile(spec_minimal)
+def test_set_version_and_release(specfile_factory, spec_minimal, version, release):
+    spec = specfile_factory(spec_minimal)
     spec.set_version_and_release(version, release)
     assert spec.version == version
     assert spec.release == release
@@ -266,8 +260,8 @@ def test_set_version_and_release(spec_minimal, version, release):
         ("patch3.patch", 3, "patch3"),
     ],
 )
-def test_add_patch(spec_autosetup, location, number, comment):
-    spec = Specfile(spec_autosetup)
+def test_add_patch(specfile_factory, spec_autosetup, location, number, comment):
+    spec = specfile_factory(spec_autosetup)
     if number == 0 or location == "patch2.patch":
         with pytest.raises(SpecfileException):
             spec.add_patch(location, number, comment)
@@ -284,8 +278,8 @@ def test_add_patch(spec_autosetup, location, number, comment):
                 assert sections.package[-4] == f"# {comment}"
 
 
-def test_remove_patches(spec_commented_patches):
-    spec = Specfile(spec_commented_patches)
+def test_remove_patches(specfile_factory, spec_commented_patches):
+    spec = specfile_factory(spec_commented_patches)
     with spec.patches() as patches:
         del patches[1:3]
         patches.remove_numbered(5)
@@ -320,8 +314,8 @@ def test_remove_patches(spec_commented_patches):
         ("%{obsrel}.%{autorelease}", True),
     ],
 )
-def test_autorelease(spec_rpmautospec, raw_release, has_autorelease):
-    spec = Specfile(spec_rpmautospec)
+def test_autorelease(specfile_factory, spec_rpmautospec, raw_release, has_autorelease):
+    spec = specfile_factory(spec_rpmautospec)
     spec.raw_release = raw_release
     assert spec.has_autorelease == has_autorelease
 
@@ -330,9 +324,9 @@ def test_autorelease(spec_rpmautospec, raw_release, has_autorelease):
     rpm.__version__ < "4.16", reason="%autochangelog requires rpm 4.16 or higher"
 )
 def test_autochangelog(
-    spec_rpmautospec, spec_conditionalized_changelog, spec_autosetup
+    specfile_factory, spec_rpmautospec, spec_conditionalized_changelog, spec_autosetup
 ):
-    spec = Specfile(spec_rpmautospec)
+    spec = specfile_factory(spec_rpmautospec)
     assert spec.has_autochangelog
     with spec.changelog() as changelog:
         assert len(changelog) == 0
@@ -341,7 +335,7 @@ def test_autochangelog(
     spec.add_changelog_entry("test")
     with spec.sections() as sections:
         assert sections.changelog == changelog
-    spec = Specfile(spec_conditionalized_changelog)
+    spec = specfile_factory(spec_conditionalized_changelog)
     assert spec.has_autochangelog
     with spec.sections() as sections:
         changelog = sections.changelog.copy()
@@ -352,7 +346,7 @@ def test_autochangelog(
     assert changelogs[0] == changelog
     with spec.changelog(changelogs[1]) as changelog:
         assert changelog[-1].content == ["test"]
-    spec = Specfile(spec_autosetup)
+    spec = specfile_factory(spec_autosetup)
     with spec.changelog() as changelog:
         changelog[0].content += "%"
     assert not spec.has_autochangelog
@@ -362,8 +356,8 @@ def test_autochangelog(
     rpm.__version__ < "4.16",
     reason="condition expression evaluation requires rpm 4.16 or higher",
 )
-def test_update_tag(spec_macros):
-    spec = Specfile(spec_macros)
+def test_update_tag(specfile_factory, spec_macros):
+    spec = specfile_factory(spec_macros)
     spec.update_tag("Version", "1.2.3~beta4")
     with spec.macro_definitions() as md:
         assert md.majorver.body == "1"
@@ -439,7 +433,7 @@ def test_update_tag(spec_macros):
         assert md.minorver.body == "2"
     with spec.sources() as sources:
         assert sources[1].location == "tests-86.tar.xz"
-    spec = Specfile(spec_macros, macros=[("use_snapshot", "1")])
+    spec = specfile_factory(spec_macros, macros=[("use_snapshot", "1")])
     spec.update_tag("Version", "3.2.1")
     with spec.macro_definitions() as md:
         assert md.majorver.body == "0"
@@ -450,7 +444,7 @@ def test_update_tag(spec_macros):
         assert md.get("package_version", 13).body == "%{mainver}%{?prever:~%{prever}}"
         assert md.get("package_version", 15).body == "3.2.1"
     assert spec.version == "%{package_version}"
-    spec = Specfile(spec_macros)
+    spec = specfile_factory(spec_macros)
     spec.update_tag("Version", "1.2.3.4~rc5")
     with spec.macro_definitions() as md:
         assert md.majorver.body == "1.2"
@@ -459,7 +453,7 @@ def test_update_tag(spec_macros):
         assert md.mainver.body == "%{majorver}.%{minorver}.%{patchver}"
         assert md.prever.body == "rc5"
     assert spec.version == "%{package_version}"
-    spec = Specfile(spec_macros)
+    spec = specfile_factory(spec_macros)
     with spec.macro_definitions() as md:
         md.prever.commented_out = True
     assert spec.expanded_version == "0.1.2"
@@ -474,9 +468,9 @@ def test_update_tag(spec_macros):
     assert spec.version == "%{package_version}"
 
 
-def test_multiple_instances(spec_minimal, spec_autosetup):
+def test_multiple_instances(specfile_factory, spec_minimal, spec_autosetup):
     spec1 = Specfile(spec_minimal)
-    spec2 = Specfile(spec_autosetup)
+    spec2 = specfile_factory(spec_autosetup)
     spec1.version = "14.2"
     assert spec2.expanded_version == "0.1"
     with spec2.sources() as sources:
@@ -485,8 +479,8 @@ def test_multiple_instances(spec_minimal, spec_autosetup):
         assert sources[1].expanded_location == "tests-0.1.tar.xz"
 
 
-def test_includes(spec_includes):
-    spec = Specfile(spec_includes)
+def test_includes(specfile_factory, spec_includes):
+    spec = specfile_factory(spec_includes)
     assert not spec.tainted
     with spec.patches() as patches:
         assert not patches
@@ -503,8 +497,8 @@ def test_includes(spec_includes):
     for inc in ["patches.inc", "provides.inc", "description1.inc", "description2.inc"]:
         (spec.sourcedir / inc).unlink()
     with pytest.raises(RPMException):
-        spec = Specfile(spec_includes)
-    spec = Specfile(spec_includes, force_parse=True)
+        spec = specfile_factory(spec_includes)
+    spec = specfile_factory(spec_includes, force_parse=True)
     assert spec.tainted
     with spec.patches() as patches:
         assert not patches
@@ -518,18 +512,18 @@ def test_includes(spec_includes):
     for inc in ["macros1.inc", "macros2.inc"]:
         (spec.sourcedir / inc).unlink()
         with pytest.raises(RPMException):
-            spec = Specfile(spec_includes, force_parse=True)
+            spec = specfile_factory(spec_includes, force_parse=True)
         assert not (spec.sourcedir / inc).is_file()
 
 
-def test_shell_expansions(spec_shell_expansions):
-    spec = Specfile(spec_shell_expansions)
+def test_shell_expansions(specfile_factory, spec_shell_expansions):
+    spec = specfile_factory(spec_shell_expansions)
     assert spec.expanded_version == "1035.4200"
     assert "C.UTF-8" in spec.expand("%numeric_locale")
 
 
-def test_context_management(spec_autosetup, spec_traditional):
-    spec = Specfile(spec_autosetup)
+def test_context_management(specfile_factory, spec_autosetup, spec_traditional):
+    spec = specfile_factory(spec_autosetup)
     with spec.tags() as tags:
         tags.license.value = "BSD"
         assert spec.license == "BSD"
@@ -541,7 +535,7 @@ def test_context_management(spec_autosetup, spec_traditional):
     assert spec.license == "BSD-3-Clause"
     with spec.patches() as patches:
         assert patches[0].location == "patch_0.patch"
-    spec1 = Specfile(spec_autosetup)
+    spec1 = specfile_factory(spec_autosetup)
     spec2 = Specfile(spec_traditional)
     with spec1.sections() as sections1, spec2.sections() as sections2:
         assert sections1 is not sections2
@@ -550,8 +544,8 @@ def test_context_management(spec_autosetup, spec_traditional):
         assert tags1 == tags2
 
 
-def test_copy(spec_autosetup):
-    spec = Specfile(spec_autosetup)
+def test_copy(specfile_factory, spec_autosetup):
+    spec = specfile_factory(spec_autosetup)
     shallow_copy = copy.copy(spec)
     assert shallow_copy == spec
     assert shallow_copy is not spec
@@ -564,7 +558,7 @@ def test_copy(spec_autosetup):
     assert deep_copy._parser is not spec._parser
 
 
-def test_parse_if_necessary(spec_macros):
+def test_parse_if_necessary(specfile_factory, spec_macros):
     flexmock(SpecParser).should_call("_do_parse").once()
     spec1 = Specfile(spec_macros)
     spec2 = copy.deepcopy(spec1)
@@ -582,12 +576,12 @@ def test_parse_if_necessary(spec_macros):
     assert spec1.expanded_version == "28.1.2~rc2"
     flexmock(SpecParser).should_receive("id").and_return(12345)
     flexmock(SpecParser).should_call("_do_parse").once()
-    spec = Specfile(spec_macros)
+    spec = specfile_factory(spec_macros)
     flexmock(SpecParser).should_call("_do_parse").never()
     assert spec.expanded_name == "test"
     spec = None
     flexmock(SpecParser).should_call("_do_parse").once()
-    spec = Specfile(spec_macros)
+    spec = specfile_factory(spec_macros)
     flexmock(SpecParser).should_call("_do_parse").never()
     assert spec.expanded_name == "test"
 
@@ -597,9 +591,9 @@ def test_parse_if_necessary(spec_macros):
     reason="condition expression evaluation requires rpm 4.16 or higher",
 )
 def test_update_version(
-    spec_prerelease, spec_prerelease2, spec_conditionalized_version
+    specfile_factory, spec_prerelease, spec_prerelease2, spec_conditionalized_version
 ):
-    spec = Specfile(spec_prerelease)
+    spec = specfile_factory(spec_prerelease)
     prerelease_suffix_pattern = r"(-)rc\d+"
     prerelease_suffix_macro = "prerel"
     spec.update_version("0.1.2", prerelease_suffix_pattern, prerelease_suffix_macro)
@@ -620,7 +614,7 @@ def test_update_version(
         assert md.prerel.body == "rc1"
         assert not md.prerel.commented_out
     assert spec.version == "%{pkgver}"
-    spec = Specfile(spec_prerelease)
+    spec = specfile_factory(spec_prerelease)
     with spec.macro_definitions() as md:
         md.prerel.commented_out = True
     spec.update_version("0.1.3-rc1", prerelease_suffix_pattern)
@@ -632,7 +626,7 @@ def test_update_version(
         assert md.prerel.body == "rc2"
         assert md.prerel.commented_out
     assert spec.version == "%{pkgver}"
-    spec = Specfile(spec_prerelease2)
+    spec = specfile_factory(spec_prerelease2)
     prerelease_suffix_pattern = r"(-)rc\d+"
     prerelease_suffix_macro = "prerel"
     spec.update_version("0.1.2", prerelease_suffix_pattern, prerelease_suffix_macro)
@@ -653,7 +647,7 @@ def test_update_version(
         assert md.prerel.body == "rc1"
         assert not md.prerel.commented_out
     assert spec.version == "%{pkgver}"
-    spec = Specfile(spec_prerelease2)
+    spec = specfile_factory(spec_prerelease2)
     with spec.macro_definitions() as md:
         md.prerel.commented_out = True
     spec.update_version("0.1.3-rc1", prerelease_suffix_pattern)
@@ -665,7 +659,7 @@ def test_update_version(
         assert md.prerel.body == "rc1"
         assert not md.prerel.commented_out
     assert spec.version == "%{pkgver}"
-    spec = Specfile(spec_conditionalized_version)
+    spec = specfile_factory(spec_conditionalized_version)
     version = "0.1.3"
     assert spec.version == "%{upstream_version}"
     spec.update_version(version, prerelease_suffix_pattern)
@@ -673,7 +667,7 @@ def test_update_version(
         assert md.upstream_version.body == version
     assert spec.version == "%{upstream_version}"
     assert spec.expanded_version == version
-    spec = Specfile(spec_conditionalized_version)
+    spec = specfile_factory(spec_conditionalized_version)
     with spec.macro_definitions() as md:
         md.commit.commented_out = False
     assert spec.version == "%{upstream_version}^git%{shortcommit}"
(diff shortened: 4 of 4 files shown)


nforro:
Thank you for the contribution!
However, I think this problem could be handled in a more elegant way, without using temporary files, and together with https://github.com/packit/specfile/issues/206. The idea is to replace the internal `Specfile._path` attribute with a file-like object that could be either passed directly to the constructor or instantiated from a path using `open()` or from a string using `StringIO`.
The implementation would simply replace instances of `Specfile._path.read_text()` with something like `Specfile._stream.read()` and instances of `Specfile._path.write_text()` with something like `Specfile._stream.write()` if the file-like object is writable. It would also need to handle seeking, truncating, flushing etc. accordingly.

p0tat0chip:
I tried implementing as described and converted even the path to a fileobject (`self._stream = self._path.open("r+")`)
But two tests were failing:
```
FAILED tests/integration/test_specfile.py::test_copy - TypeError: cannot pickle 'TextIOWrapper' instances
FAILED tests/integration/test_specfile.py::test_parse_if_necessary - TypeError: cannot pickle 'TextIOWrapper' instances
```

nforro:
Right, that will require doing something similar to what is done in `SpecParser`:

https://github.com/packit/specfile/blob/01f6a7c8e8c1ec45ef0deefe3883afe3de73bcfb/specfile/spec_parser.py#L81-L90

The `_stream` attribute will have to be skipped when deepcopying a `Specfile` instance, but the question is what should the copy get. For `Specfile` instantiated from a path it could be calling `open()` again on the same path, for `Specfile` instantiated from a string creating a new `StringIO` instance and supplying it with the content of the original `StringIO`. For `Specfile` instantiated with a file-like object perhaps we could just disable deepcopying.

p0tat0chip:
The tests are passing now. Currently:
- Return a new `StringIO` instance for the ones instantiated from it.
- Return a copy for the ones instantiated from `path`. They are still file-like objects from `open()`.
- Ignore and return None for file-like objects without `path`.

specfile/specfile.py
```diff
@@ -71,9 +84,27 @@ def __init__(
...
+            self._path = Path(path)
+            self._stream = self._path.open(
+                "r+", encoding="utf8", errors="surrogateescape"
+            )
+        else:
+            raise ValueError("Either 'stream' or 'path' must be provided")
+
+        self._is_writable = hasattr(self._stream, "write") and callable(
+            getattr(self._stream, "write", None)
+        )
```
nforro, line 105: I would drop this attribute, as well as `_path`, both can be inferred from `_stream`. You can make it a property if you think it would be useful. The `path` property should definitely stay for backwards compatibility, but its setter must be adjusted - if a user changes path, `_stream` has to be updated/replaced.

specfile/specfile.py
```diff
@@ -52,7 +64,8 @@ class Specfile:
 
     def __init__(
         self,
-        path: Union[Path, str],
+        path: Optional[Union[Path, str]] = None,
+        stream: Optional[StringIO] = None,
```
nforro, line 68: I would probably use [ `file`](https://docs.python.org/3/glossary.html#term-file-object) as the argument name rather than `stream`, I think it's a bit more self-explanatory. Its meaning should definitely be documented in the docstring. The type should be `IOBase`, to cover all possibilities. There should be another argument of type `str` that, if specified, will create a `StringIO` instance and assign it to the `_stream` attribute (or perhaps `_file` to keep the naming consistent).

nforro (changes requested):
Nice, this looks much better.

p0tat0chip:
I had a doubt in the `reload` function:
```python
    def reload(self) -> None:
        """Reloads the spec file content."""
        self._lines, self._trailing_newline = self._read_lines(self._file)
```
Why we are not parsing again here ?  I think this causes the object to be inconsistent after a reload.

nforro:
> Why we are not parsing again here ? I think this causes the specfile object to be inconsistent after a reload.

By parsing do you mean calling `self._parser.parse()`? Because it would be unnecessary, it is called before any data that depend on the result of RPM parsing are accessed.

p0tat0chip:
> > Why we are not parsing again here ? I think this causes the specfile object to be inconsistent after a reload.
> 
> By parsing do you mean calling `self._parser.parse()`? Because it would be unnecessary, it is called before any data that depend on the result of RPM parsing are accessed.

yes, I was talking about that. It's clear now

specfile/specfile.py
```diff
@@ -120,18 +183,32 @@ def _dump_debug_info(self, message) -> None:
         )
 
     @staticmethod
-    def _read_lines(path: Path) -> Tuple[List[str], bool]:
-        content = path.read_text(encoding="utf8", errors="surrogateescape")
-        return content.splitlines(), content[-1] == "\n"
+    def _read_lines(stream: TextIO) -> Tuple[List[str], bool]:
+        stream.seek(0)
+        content = stream.read()
```
nforro, line 188: The argument doesn't have to be `TextIO` (I'm surprised mypy didn't catch this):

```suggestion
    def _read_lines(file: IOBase) -> Tuple[List[str], bool]:
        file.seek(0)
        raw_content = file.read()
        if isinstance(raw_content, str):
            content = raw_content
        else:
            content = raw_content.decode("utf8", errors="surrogateescape")
```

Question is what to do with custom `TextIOBase`-based objects as we can't control their `encoding` and `errors` attributes. I suppose it's a responsibility of the user to make sure the spec file content can be properly decoded. So we should document that.

specfile/specfile.py
```diff
@@ -71,12 +87,37 @@ def __init__(
...
+            self._file = StringIO(raw_string)
+            if sourcedir is None:
+                raise ValueError(
+                    "'sourcedir' must be defined when providing string input"
+                )
+        else:
+            raise ValueError(
+                "Either 'file', 'path', or 'string_input' must be provided"
+            )
+        self.is_sourcedir_set = sourcedir is not None
```
nforro, line 112: What's the point of this? Why evaluate a condition at the time of instantiation and then use the result later when it doesn't have to be valid anymore? Also, there is already the `sourcedir` property.

specfile/specfile.py
```diff
@@ -120,18 +183,32 @@ def _dump_debug_info(self, message) -> None:
...
+    def _read_lines(stream: TextIO) -> Tuple[List[str], bool]:
+        stream.seek(0)
+        content = stream.read()
+        return content.splitlines(), content.endswith("\n")
+
+    @property
+    def is_writable(self) -> bool:
+        return hasattr(self._file, "write") and callable(
+            getattr(self._file, "write", None)
+        )
```
nforro, line 195: Still not sure how useful this will be, but:
- let's keep the naming consistent, there is already e.g. `tainted` property, not `is_tainted`
- `IOBase` has `writable()` method: https://docs.python.org/3/library/io.html#io.IOBase.writable

```suggestion
    def writable(self) -> bool:
        return self._file.writable()
```

specfile/specfile.py
```diff
@@ -120,18 +183,32 @@ def _dump_debug_info(self, message) -> None:
...
 
     @property
     def path(self) -> Path:
         """Path to the spec file."""
-        return self._path
+        if hasattr(self._file, "name"):
+            return Path(self._file.name)
+        return None
```
nforro, line 202:
```suggestion
        try:
            return Path(self._file.name)
        except AttributeError:
            return None
```

Also, the return type is wrong, it should be `Optional[Path]`.

specfile/specfile.py
```diff
@@ -120,18 +183,32 @@ def _dump_debug_info(self, message) -> None:
...
 
     @path.setter
     def path(self, value: Union[Path, str]) -> None:
-        self._path = Path(value)
+        self._file.close()
```
nforro, line 206: This should be done implicitly upon destruction.

specfile/specfile.py
```diff
@@ -120,18 +183,32 @@ def _dump_debug_info(self, message) -> None:
...
+        new_path = Path(value)
+        self._file = new_path.open("r+", encoding="utf8", errors="surrogateescape")
```
nforro, line 208: Since the same arguments are used on multiple places, it would be nice to make them a constant, so they can be changed at a single location (if needed).

specfile/specfile.py
```diff
@@ -120,18 +183,32 @@ def _dump_debug_info(self, message) -> None:
...
+        if not self.is_sourcedir_set:
+            self._parser.sourcedir = new_path.parent
+        self.reload()
```
nforro, line 211: I wouldn't do this, the original implementation also didn't change sourcedir when changing path. Even if a user didn't specify sourcedir explicitly (which you can't possibly know for sure from constructor arguments, the user could use the sourcedir property to do that), they may want to keep the sources where they are and just move the spec file elsewhere. It is true though that we should document this - if they change path sourcedir won't be automatically updated even if they didn't specify it explicitly.

specfile/specfile.py
```diff
@@ -179,11 +256,16 @@ def rpm_spec(self) -> rpm.spec:
...
     def reload(self) -> None:
         """Reloads the spec file content."""
-        self._lines, self._trailing_newline = self._read_lines(self.path)
+        self._lines, self._trailing_newline = self._read_lines(self._file)
 
     def save(self) -> None:
         """Saves the spec file content."""
-        self.path.write_text(str(self), encoding="utf8", errors="surrogateescape")
+        if not self.is_writable:
+            raise OSError("This stream is read-only")
```
nforro, line 264: I think this is unnecessary, a (perhaps more useful) exception would be raised anyway.

specfile/specfile.py
```diff
@@ -179,11 +256,16 @@ def rpm_spec(self) -> rpm.spec:
...
+        self._file.seek(0)
+        self._file.truncate(0)
+        self._file.write(str(self))
```
nforro, line 267: Again, `self._file` doesn't have to be text-based:

```suggestion
        content = str(self)
        try:
            self._file.write(content)
        except TypeError:
            self._file.write(content.encode("utf8", errors="surrogateescape"))
```

nforro (changes requested):

specfile/specfile.py
```diff
@@ -71,12 +87,37 @@ def __init__(
                 Such sources include sources referenced from shell expansions
                 in tag values and sources included using the _%include_ directive.
         """
+        if file is not None:
+            self._file = file
+            if hasattr(file, "name") and file.name and sourcedir is None:
+                raise ValueError(
+                    "'sourcedir' must be defined when providing a file with a name"
```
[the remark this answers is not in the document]
p0tat0chip: Oh, sorry i was testing some things and forgot to remove the hasattr()

specfile/specfile.py
p0tat0chip: Right, seems like I got extra paranoid with the checks

specfile/specfile.py
p0tat0chip: I was using it to determine whether to set sourcedir explicitly, which was used to determine how to handle changes of `path`. I will remove it seems like i misunderstood.

specfile/specfile.py
p0tat0chip: I will just remove it for now

specfile/specfile.py
nforro: I see, but like I said in some other comment, `sourcedir` can be changed outside of the constructor. While it would be possible to keep the attribute and update it also in `sourcedir` setter, I'm not sure it's worth it. If you want to keep it I would make it private and give it a more self-explanatory name, something like `_sourcedir_set_explicitly`.

p0tat0chip: I think i should just remove it and keep things like before, since it's complicating it.

specfile/specfile.py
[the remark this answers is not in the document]
nforro: My question still stands - why must be sourcedir defined when providing a named file but not when providing a path?

specfile/constants.py
```diff
@@ -244,3 +244,5 @@
     "x86_64_v4",
     "xtensa",
 }
+
+FILE_RW = {"mode": "r+", "encoding": "utf8", "errors": "surrogateescape"}
```
nforro, line 248: Sorry, I meant the `encoding` and `errors` arguments that are shared among all `encode()`/`decode()`/`open()` methods. I would also make it a `Specfile`-specific constant, or at least move it to `specfile.py`, this file is intended to hold RPM constants and is supposed to be updated programatically. I would probably call the constant something like `ENCODING`, the `errors` argument is tightly tied to encoding.

nforro (commented):
It's starting to look really good, thanks!
Could you please squash the commits into one (or a couple if you feel it's necessary)?
It would also be nice to check why the source-git tests in packit integration test suite are failing and if it is related to these changes.

p0tat0chip:
```
FAILED integration/test_source_git.py::test_basic_local_update_without_patching
FAILED integration/test_source_git.py::test_basic_local_update_empty_patch[None]
FAILED integration/test_source_git.py::test_basic_local_update_empty_patch[0.1.0]
FAILED integration/test_source_git.py::test_basic_local_update_empty_patch[0.1*]
FAILED integration/test_source_git.py::test_basic_local_update_empty_patch[0.*]
FAILED integration/test_source_git.py::test_basic_local_update_patch_content
FAILED integration/test_source_git.py::test_basic_local_update_patch_content_with_metadata
FAILED integration/test_source_git.py::test_basic_local_update_patch_content_with_metadata_and_patch_ignored
FAILED integration/test_source_git.py::test_basic_local_update_patch_content_with_downstream_patch
FAILED integration/test_source_git_status.py::test_source_git_status_synced[check_ready_api_dg_first]
FAILED integration/test_source_git_status.py::test_source_git_status_synced[check_ready_api_sg_first]
FAILED integration/test_source_git_status.py::test_source_git_status_dist_git_ahead[check_ready_api_dg_first]
FAILED integration/test_source_git_status.py::test_source_git_status_dist_git_ahead[check_ready_api_sg_first]
FAILED integration/test_source_git_status.py::test_source_git_status_source_git_ahead[check_ready_api_dg_first]
FAILED integration/test_source_git_status.py::test_source_git_status_source_git_ahead[check_ready_api_sg_first]
FAILED integration/test_source_git_status.py::test_source_git_status_history_diverges[check_ready_api_dg_first]
FAILED integration/test_source_git_status.py::test_source_git_status_history_diverges[check_ready_api_sg_first]
FAILED integration/test_upstream.py::test_fix_spec_persists - AssertionError:...
```

These are the failing tests, I think the last one might have something to do with this PR

specfile/specfile.py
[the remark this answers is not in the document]
p0tat0chip: I used that condition taking a note from here https://github.com/packit/specfile/issues/206#issuecomment-1418672889 .
I misinterpreted.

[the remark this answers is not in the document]
nforro: > sourcedir should now be required only in case of strings only, where it's not possible to infer a path.

No, it should be required when `_file` doesn't have a name, that's the case for way more classes than just `StringIO`.

nforro:
> These are the failing tests, I think the last one might have something to do with this PR

No, all of the failures are related to this PR. I haven't check all of them but they all most likely fail for the same reason - the spec file is replaced with a different file, so the `_file` object represents a deleted file and `reload()` doesn't read contents of the new file. I think that's fine, but it will be necessary to add a new method, called for example `reopen()`, that will reopen the path, but do nothing on other types of file-like objects.

nforro:
> I am having trouble squashing commits from my side they seem to be getting duplicated.

https://medium.com/@slamflipstrom/a-beginners-guide-to-squashing-commits-with-git-rebase-8185cf6e62ec

nforro:
> I think that's fine, but it will be necessary to add a new method, called for example `reopen()`, that will reopen the path, but do nothing on other types of file-like objects.

Actually, thinking about it, it should rather be done in `reload()`, in order not to change the behavior for existing users.

p0tat0chip:
> > I am having trouble squashing commits from my side they seem to be getting duplicated.
> 
> https://medium.com/@slamflipstrom/a-beginners-guide-to-squashing-commits-with-git-rebase-8185cf6e62ec

Got it done eventually i was avoiding force push and rewriting the commits.
Yes, thinking about it, the issue seems to be with reload. There was an assertion error in a test, where it still had the old version string.

p0tat0chip:
> > These are the failing tests, I think the last one might have something to do with this PR
> 
> No, all of the failures are related to this PR. I haven't check all of them but they all most likely fail for the same reason - the spec file is replaced with a different file, so the `_file` object represents a deleted file and `reload()` doesn't read contents of the new file. I think that's fine, but it will be necessary to add a new method, called for example `reopen()`, that will reopen the path, but do nothing on other types of file-like objects.
> 
> That brings me to another step that will be necessary and that's adding tests covering every new feature and changed behavior, including this one.

I think the current test cases should be reused, but with different input types, such as a raw string, a file-based input, and a file without a 'name' attribute.

nforro:
> it might be overkill though

Quite the opposite, every supported case should be covered with tests.

nforro:
> happens since the file_path to be reopened does not exist anymore after `close()`

Hm, that doesn't add up, same thing would have happened with the original implementation and `self.path.read_text()`. I'll take a look.

nforro:
@p0tat0chip you can [install pre-commit](https://github.com/packit/contributing?tab=readme-ov-file#pre-commit) and it will automatically reformat the code for you to match the coding style of this project.

specfile/specfile.py
```diff
@@ -120,18 +189,28 @@ def _dump_debug_info(self, message) -> None:
...
+            return Path(self._file.name)
+        except AttributeError:
+            return None
 
     @path.setter
     def path(self, value: Union[Path, str]) -> None:
-        self._path = Path(value)
+        self._file.close()
+        new_path = Path(value)
+        self._file = new_path.open("r+", **ENCODING_ARGS)
```
nforro, line 213: There are actually a few possible scenarios and questions here:
- `Specfile` has been instantiated with a file or a string and the user is now setting `path`
  - do we want to allow this at all?
  - if yes, should we just silently open a new file? should we close the old one?
  - what if `self._file.name == value`?
- `Specfile` has been instantiated with a path
  - do we want to open a new file even if the path hasn't changed?

specfile/specfile.py
```diff
@@ -179,11 +258,26 @@ def rpm_spec(self) -> rpm.spec:
...
-        self._lines, self._trailing_newline = self._read_lines(self.path)
+        try:
+            if isinstance(self._file, IOBase) and hasattr(self._file, "name"):
+                file_path = Path(self._file.name)
+                if file_path.exists():
+                    self._file.close()
+                    self._file = file_path.open("r+", **ENCODING_ARGS)
+            self._lines, self._trailing_newline = self._read_lines(self._file)
+        except AttributeError:
+            self._lines, self._trailing_newline = self._read_lines(self._file)
```
nforro, line 269: There are quite some unnecessary checks - `isinstance(self._file, IOBase)` is always true, `hasattr(self._file, "name")` is covered by the `try`-`except`. As for checking the existence, it doesn't make much sense, if the path doesn't exist anymore we do nothing?

```suggestion
        try:
            path = Path(self._file.name)
        except AttributeError:
            pass
        else:
            # reopen the file, in case it has been replaced
            self._file.close()
            self._file = path.open("r+", **ENCODING_ARGS)
        self._lines, self._trailing_newline = self._read_lines(self._file)
```

tests/integration/test_specfile.py
```diff
@@ -15,19 +15,30 @@
...
-    for path in spec.sourcedir.iterdir():
-        if not path.samefile(spec.path):
-            path.unlink()
-    spec = Specfile(spec_multiple_sources)
-    assert spec.rpm_spec.prep == prep
-
-
-def test_prep_traditional(spec_traditional):
-    spec = Specfile(spec_traditional)
+def create_specfile(input_path, mode="path", **kwargs):
```
nforro, line 18: Doing this on every test in unnecessary, but why not. However, instead of parametrizing every test function, make this a fixture, parametrize it and use it in the test functions.

tests/integration/test_specfile.py
```diff
@@ -15,19 +15,30 @@
...
+
+    :param input_path: Path to the spec file.
+    :param mode: Determines how to pass input ("path", "file", "raw_string").
+    :return: Specfile instance.
+    """
+    if mode == "path":
+        return Specfile(input_path, **kwargs)
+    elif mode == "file":
+        with open(input_path, "r+", encoding="utf-8", errors="surrogateescape") as f:
+            return Specfile(file=f, **kwargs)
```
nforro, line 30: I would add at least `StringIO`, `BytesIO` and `BufferedRandom` (`open(mode="rb+")`) to the list.

nforro (changes requested):
Very nice! I think we are only missing a test for the reload after the spec file has been replaced with another file.

specfile/specfile.py
```diff
@@ -71,12 +88,35 @@ def __init__(
...
+        elif raw_string is not None:
+            self._file = StringIO(raw_string)
+            if sourcedir is None:
+                raise ValueError(
+                    "'sourcedir' is required when providing a raw string input"
+                )
+        else:
+            raise ValueError(
+                "Either 'file', 'path', or 'string_input' must be provided"
+            )
```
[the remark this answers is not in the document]
p0tat0chip: > I think this needs a bit more work. What if someone specifies e.g. both `file` and `path`? The arguments should be mutually exclusive and an exception should be raised if more than one is specified, or their priority must be clearly documented - i.e. `if file is specified, path and raw_string are ignored` etc.
> 
> I'm also not sure about the `raw_string`, in Python the term is used to represent the `r"some_str"` string literal, and it's not very self-explanatory in my opinion, what do you think about using `content` instead?

That's more sensible name, will do that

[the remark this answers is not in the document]
p0tat0chip: I think something like this should suffice for mutual exclusivity and just add the information in docstrings.

```python
        
        if sum([file is not None, path is not None, content is not None]) > 1:
                    raise ValueError("Only one of 'file', 'path', or 'content' should be provided")
        if file is not None:
            self._file = file
            ....
```

specfile/specfile.py
p0tat0chip: The file.exists() check was added to ensure the previously failing cockpit test passes. If the path no longer exists and cannot be reopened, the existing file content is retained. Without this check, _file, self._lines, and self._trailing_newline would never be initialized, preventing _read_lines(self._file) from executing.
I do wonder if this was even remotely a valid fix.

tests/integration/test_specfile.py
p0tat0chip: Alright, this is clearly more elegant.

specfile/specfile.py
nforro: I wonder, is it the same test I have fixed [here](https://github.com/packit/packit/pull/2532/files)? The PR is blocked on something else so it's not merged yet, but if the fix works in this case we can merge it separately.

p0tat0chip: The previous implementation did not reopen any file handles and simply read again from the already opened file.
To preserve that behaviour, I just read the already open file again and do nothing.
What do you think if it was fine to do so ? I am a bit puzzled.

nforro: > The previous implementation did not reopen any file handles and simply read again from the already opened file.

Yes, and that didn't match the original behavior so it was changed. There should be no exceptions to that because of a wrongly written test (if that's the case).

nforro: Thinking about it, [this](https://github.com/packit/packit/pull/2564) could also fix the test (and is probably a reason why it worked in the first place, though I still think it is badly written).

tests/integration/test_specfile.py
p0tat0chip: I added checks for the encoding attribute for deepcopy in case of BufferedRandom since it lacks the support of

specfile/specfile.py
```diff
@@ -111,6 +156,44 @@ def __exit__(
...
+                    specfile._file = path.open(self._file.mode)
+                else:
+                    specfile._file = path.open(
+                        self._file.mode, encoding=encoding, errors=errors
+                    )
+        else:
+            if hasattr(self._file, "encoding"):
+                specfile._file = type(self._file)(data)
+            else:
+                specfile._file = BytesIO(data)
```
nforro, line 193: What's the point of this?

specfile/specfile.py
```diff
@@ -179,11 +272,25 @@ def rpm_spec(self) -> rpm.spec:
 
     def reload(self) -> None:
         """Reloads the spec file content."""
-        self._lines, self._trailing_newline = self._read_lines(self.path)
+        try:
+            path = Path(self._file.name)
+        except AttributeError:
+            pass
+        else:
```
nforro, line 279:
```suggestion
        else:
            # reopen the path in case the original file has been deleted/replaced
```

tests/integration/conftest.py
```diff
@@ -136,3 +138,44 @@ def spec_conditionalized_version(tmp_path):
     specfile_path = tmp_path / SPECFILE
     shutil.copyfile(SPEC_CONDITIONALIZED_VERSION / SPECFILE, specfile_path)
     return specfile_path
+
+
+@pytest.fixture(
+    params=["path", "file", "raw_string", "stringio", "bytesio", "bufferedrandom"]
```
nforro, line 144: Why not use `text_file` and `binary_file` rather than (now ambiguous) `file` and type name? Also, I would not make the type names lowercase. Or you could do the same as with `file` and use `text_io` and `binary_io`.

specfile/specfile.py
p0tat0chip: it's mostly redundant, that line was used to handle binary and text based file objects, binary ones don't have the encoding attribute. I will remove it :sweat_smile:

p0tat0chip: I was thinking for the cases of BytesIO, StringIO and BufferedRandom should we keep the error about not supporting deep-copy or just a return a new instance. So the error message is even needed now ?

nforro: Most, but not necessarily all. Anyone can create a new class derived from e.g. `TextIOBase` that doesn't have `path` nor `getvalue()`.

p0tat0chip:
```python
        try:
            specfile._file = deepcopy(self._file.getvalue())
        except AttributeError:
            if self._file.name:
                path = Path(self._file.name)
                mode = self._file.mode
                if getattr(self._file, "encoding", None):
                    specfile._file = path.open(
                        mode, encoding=self._file.encoding, errors=self._file.errors
                    )
                else:
                    specfile._file = path.open(mode)
            else:
                raise ValueError(
                    "Deep copy not supported for arbitrary file-like objects"
                )
```

This should be good.

nforro: Is that a pseudo code? :slightly_smiling_face: The second line makes no sense, the fourth can raise `AttributeError` that's not handled, and I don't understand the tendency to prefer `getattr` over more pythonic `try`-`except`.

p0tat0chip: try-except was raising errors about the missing encoding attribute in some other case, I left it there oops.

p0tat0chip:
```python
try:
    specfile._file = deepcopy(self._file.getvalue())
except AttributeError:
    try:
        path = Path(self._file.name)
        mode = self._file.mode
        try:
            specfile._file = path.open(
                mode, encoding=self._file.encoding, errors=self._file.errors
            )
        except AttributeError:
            specfile._file = path.open(mode)
    except AttributeError:
        raise ValueError("Deep copy not supported for arbitrary file-like objects")
```

p0tat0chip:
```python
        try:
            # Try copying in-memory file-like objects (StringIO, BytesIO)
            specfile._file = type(self._file)(self._file.getvalue())
        except AttributeError:
            try:
                # handle named file objects
                specfile._file = Path(self._file.name).open(
                    mode=self._file.mode,
                    encoding=getattr(self._file, "encoding", None),
                    errors=getattr(self._file, "errors", None),
                )
            except AttributeError:
                # Handle BufferedRandom and other file-like objects without getvalue()
                specfile._file = type(self._file)()
                specfile._file.write(self._file.read())  # Copy contents
                specfile._file.seek(0)
```

There shouldn't be any sins left now.

tests/integration/test_specfile.py
```diff
@@ -684,8 +668,21 @@ def test_update_version(
...
+    spec = specfile_factory(spec_minimal)
+    assert spec.version == "0.1"
+
+    # Simulate an external file change by replacing StringIO content
+    updated_content = StringIO(
+        "Name: test-package\nVersion: 2.0\nRelease: 2\nLicense: MIT\n"
+    )
+    spec._file = updated_content
+    spec.reload()
+    assert spec.version == "2.0"
```
nforro, line 688: I think this would always pass. This test case should be about removing a path to the file `Specfile` has been instantiated with and creating a new file at the same path. The `Specfile` instance shouldn't be touched, only `reload()` should be called and the new content should match the new file.

tests/integration/conftest.py
```diff
@@ -136,3 +138,44 @@ def spec_conditionalized_version(tmp_path):
     specfile_path = tmp_path / SPECFILE
     shutil.copyfile(SPEC_CONDITIONALIZED_VERSION / SPECFILE, specfile_path)
     return specfile_path
+
+
+@pytest.fixture(
+    params=["path", "file", "raw_string", "text_io", "binary_io", "bufferedrandom"]
```
nforro, line 144: Again, why `file` and `bufferedrandom`? It makes no sense to me. Why don't you use `textiowrapper` instead of `file` then? (I'm not suggesting you should to that, I'm just trying to point out why `bufferedrandom` doesn't make sense here).

specfile/specfile.py
```diff
@@ -71,12 +90,31 @@ def __init__(
...
+            self._file = Path(path).open("r+", **ENCODING_ARGS)
+        elif content is not None:
+            self._file = StringIO(content)
+        else:
+            raise ValueError(
+                "Either 'file', 'path', or 'string_input' must be provided"
+            )
+        if sourcedir is None:
+            try:
+                sourcedir = Path(self._file.name).parent
```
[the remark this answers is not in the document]
nforro: Never mind, it's too late, I should go to sleep :slightly_smiling_face:

tests/integration/conftest.py
p0tat0chip: I think `buffered_io` should be fine then.

tests/integration/test_specfile.py
p0tat0chip: And for the cases of in-memory file objects like StringIO, should the test be skipped ? or update the reload function,  In the current state, it fails to reload.

specfile/specfile.py
```diff
@@ -111,6 +149,36 @@ def __exit__(
...
+                    encoding=getattr(
+                        self._file, "encoding", None
+                    ),  # Handle attr for BufferedRandom
+                    errors=getattr(self._file, "errors", None),
+                )
+            except AttributeError:
+                # Handle BufferedRandom and other file-like objects without getvalue()
+                specfile._file = type(self._file)()
+                specfile._file.write(self._file.read())  # Copy contents
+                specfile._file.seek(0)
```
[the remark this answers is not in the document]
p0tat0chip: This block is just for the case of in-memory file-objects like StringIO and BufferedRandom, we create a new instance of their type and copy over the contents.

For the `encoding` and `error` when getattr evals to False, it returns None and returns `self._file.encoding/error` if True and does not actually throw an error. So the only AttributeError that is being catched is of `name`.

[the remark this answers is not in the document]
nforro: > This block is just for the case of in-memory file-objects like BufferedRandom, we create a new instance of their type and copy over the contents.

`BufferedRandom` is not an in-memory file object, it has the `name` attribute and you can't instantiate it without arguments.

> For the `encoding` and `error` when getattr evals to False, it returns None and returns `self._file.encoding/error` if True and does not actually throw an error.

That's my point, but ok, as `open()` accepts the `encoding`/`errors` arguments even in binary mode your way works too, though I would still argue it's less pythonic than `try`-`except`.

What I don't understand is the `BufferedRandom`, you mention it in both `try` and `except` blocks, which makes zero sense, also in the `try` block `_file` doesn't have to be `BufferedRandom` at all (and it most likely won't be), it can be e.g. `BufferedReader` or `BufferedWriter` or some other class derived from `FileIO`.

tests/integration/conftest.py
nforro: > I think `buffered_io` should be fine then.

What's the difference between `file` and `text_io` then? And is `buffered_io` text or binary? `TextIOWrapper` is buffered too.

Why not use simple (though technically still a little ambiguous) `text_file`, `binary_file`, `text_io` and `binary_io`?

tests/integration/test_specfile.py
nforro: Well, if the `Specfile` is backed by e.g. `StringIO`, there is no path to remove and recreate, but you can still test that `reload()` does the right thing after replacing the content of the file object `Specfile` has been instantiated with.

specfile/specfile.py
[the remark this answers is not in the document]
p0tat0chip: I may have misunderstood BufferedRandom , and about the point being pythonic I do agree but getattr seemed more concise, I will do the `try-except`  then.

tests/integration/conftest.py
nforro: No problem, it just seems you got focused on `BufferedRandom`, making it somehow special, while in reality it's just one of the possible return types of `open()`.

specfile/specfile.py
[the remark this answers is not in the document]
mynk8: We first check if the file is readable and seekable and save it as a boolean.
1. Check by getvalue():
	checks for `isinstance` StringIO or ByteIO, mostly for in-memory types.

2. Check by path-based reopens:
	checks for `name` and `mode` attributes, handle BufferedRandom, BufferedWriter, BufferedReader and TextIOWrapper type.

3. Fallback for check 1 and 2, handle cases of BufferedRandom/Reader/Writer, etc, lacking path or getvalue() method comes after path check to avoid falsely handling those types with path. Also the file should be seekable and readable

4. Fallback for any unknown file type:
	if the file is seekable and readable, it takes the content from it and renders a StringIO or BytesIO. This is the absolute last way to handle it.

Also, for BufferedWriter it can be instantiated from a file or not, it is handled by check 2 and 3 respectively.

Is this approach sound and safer ? Do you think it covers everything ? 
I will push the code tomorrow positively after completing tests.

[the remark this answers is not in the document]
nforro: > 2\. checks for `name` and `mode` attributes, handle BufferedRandom, BufferedWriter, BufferedReader and 
TextIOWrapper type. We can handle a type like BufferedWriter which is not readable by reopening the file from path.

Why deal with `mode`? What's wrong with accessing `encoding` and falling back to `open()` without it on `AttributeError`?
Also, if `BufferedWriter` is not readable then it will almost certainly not appear here - an exception would be raised when instantiating the original `Specfile`.

> 4\. Fallback for any unknown file type:
>   if the file is seekable and readable, it takes the content from it and renders a StringIO or BytesIO. This is the absolute last way to handle it.

I would just raise an exception in this case. I'm not even sure if 3 makes sense, you don't know what arguments the constructor could expect, but it could probably work in most cases, so I'm not against.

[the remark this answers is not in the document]
mynk8: 3 was for some very edge case i thought, what if someone decides to wrap BytesIO or a binary file in BufferedReader, for reading from a network stream. That object wouldn't have neither `name` nor `path`. 

And yes i forgot to mention for check 2 I am using `encoding/error` to check for binary types and checking for mode is unnecessary i guess.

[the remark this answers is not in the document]
nforro: > 3 was for some very edge case i thought, what if someone decides to wrap BytesIO or a binary file in BufferedReader, for reading from a network stream. That object wouldn't have neither `name` nor `path`.

Yes, but the constructor of such class can take required arguments (e.g. network related in this case). But I suppose it's ok to try and fail.

> And yes i forgot to mention for check 2 I am using `encoding/error` to check for binary types and checking for mode is unnecessary i guess.

Great :slightly_smiling_face:

tests/integration/conftest.py
```diff
@@ -136,3 +138,61 @@ def spec_conditionalized_version(tmp_path):
...
+
+        if mode == "file_path":
+            return Specfile(path=input_path, **kwargs)
+        elif mode == "text_file":
+            f = open(input_path, "r+", encoding="utf-8", errors="surrogateescape")
+            return Specfile(file=f, **kwargs)
+        elif mode == "content_string":
+            with open(input_path, encoding="utf-8", errors="surrogateescape") as f:
+                content = f.read()
+            return Specfile(content=content, **kwargs)
```
nforro, line 181:
```suggestion
            content = Path(input_path).read_text(
                encoding="utf-8", errors="surrogateescape"
            )
            return Specfile(content=content, **kwargs)
```

tests/integration/conftest.py
```diff
@@ -136,3 +138,61 @@ def spec_conditionalized_version(tmp_path):
...
+        elif mode == "text_io_stream":
+            with open(input_path, encoding="utf-8", errors="surrogateescape") as f:
+                content = f.read()
+            string_io = io.StringIO(content)
+            return Specfile(file=string_io, **kwargs)
```
nforro, line 186:
```suggestion
            content = Path(input_path).read_text(
                encoding="utf-8", errors="surrogateescape"
            )
            return Specfile(file=io.StringIO(content), **kwargs)
```

tests/integration/conftest.py
```diff
@@ -136,3 +138,61 @@ def spec_conditionalized_version(tmp_path):
...
+        elif mode == "binary_io_stream":
+            with open(input_path, "rb") as f:
+                content_bytes = f.read()
+            bytes_io = io.BytesIO(content_bytes)
+            return Specfile(file=bytes_io, **kwargs)
```
nforro, line 191:
```suggestion
            content = Path(input_path).read_bytes()
            return Specfile(file=io.BytesIO(content), **kwargs)
```

tests/integration/conftest.py
```diff
@@ -136,3 +138,61 @@ def spec_conditionalized_version(tmp_path):
...
+        elif mode == "binary_file":
+            f = open(input_path, "rb+")
+            return Specfile(file=f, **kwargs)
+        else:
+            raise ValueError(f"Unknown mode: {mode}")
```
nforro, line 196: Again, it would make more sense to me to sort the conditions as I mentioned above.

nforro, line 196: This can't possibly happen, I would just remove it.

tests/integration/test_specfile.py
```diff
@@ -684,8 +679,31 @@ def test_update_version(
...
+        after_reload = spec
+    else:
+        spec._file.seek(0)
+        spec._file.truncate()
+        with open(spec_traditional, encoding="utf-8") as f:
+            new_content = f.read()
+        try:
+            spec._file.write(new_content)
+        except TypeError:
+            spec._file.write(new_content.encode("utf-8"))
```
nforro, line 704: The `errors` argument is important, using `surrogateescape` for both decoding and encoding ensures the original binary representation will be preserved.

```suggestion
        content = Path(spec_traditional).read_text(encoding="utf-8", errors="surrogateescape")
        try:
            spec._file.write(content)
        except TypeError:
            spec._file.write(content.encode("utf-8", errors="surrogateescape"))
```

tests/integration/test_specfile.py
```diff
@@ -684,8 +679,31 @@ def test_update_version(
...
+        spec._file.flush()
+        spec._file.seek(0)
+        spec.reload()
+        after_reload = spec
+    assert str(before_reload) != str(after_reload)
```
nforro, line 709: This is nice and indeed does test all "modes", however it still doesn't test the case that caused troubles in packit. That of course makes sense only with named files, and their path needs to be removed and replaced with a different file, or, in other words, the path inode has to change.

tests/integration/conftest.py
```diff
@@ -136,3 +138,50 @@ def spec_conditionalized_version(tmp_path):
...
+    """
+    mode = request.param
+
+    def _create_specfile(input_path, **kwargs):
+        kwargs.setdefault("sourcedir", Path(input_path).parent)
+
+        if mode == "file_path":
+            return Specfile(path=input_path, **kwargs)
+        elif mode == "text_file":
+            f = open(input_path, "r+", encoding="utf-8", errors="surrogateescape")
```
nforro, line 168: You could also import `ENCODING_ARGS` from `specfile.specfile` and use them here and below. Though it would be probably a bit nicer if they were a class constant so you could just do `Specfile.ENCODING_ARGS`.

tests/integration/conftest.py
```diff
@@ -136,3 +138,50 @@ def spec_conditionalized_version(tmp_path):
...
+    """
+    pytest fixture to create a `Specfile` instance with different input modes.
+
+    Returns:
+        Function that creates a `Specfile` instance.
+    """
+    mode = request.param
+
+    def _create_specfile(input_path, **kwargs):
+        kwargs.setdefault("sourcedir", Path(input_path).parent)
```
nforro, line 163: I didn't realize `input_path` should always be a `Path`, so you could drop the `Path(input_path)` everywhere. Or, just to be 100% sure, convert it to `Path` here at the top and then use it as it is.

tests/integration/test_specfile.py
```diff
@@ -684,8 +681,35 @@ def test_update_version(
...
     assert str(spec)[-1] != "\n"
+
+
+def test_reload(specfile_factory, spec_minimal, spec_traditional):
+    spec = specfile_factory(spec_minimal)
+    before_reload = copy.deepcopy(spec)
+
+    if spec.path is not None:
+        os.remove(spec_minimal)
+        shutil.copy2(spec_traditional, spec_minimal)
```
nforro, line 697: I think this should have the same effect and you can drop the extra imports.

```suggestion
        spec_minimal.unlink()
        spec_minimal.write_bytes(spec_traditional.read_bytes())
```

tests/integration/test_specfile.py
```diff
@@ -684,8 +681,35 @@ def test_update_version(
...
+    spec = specfile_factory(spec_no_trailing_newline)
     assert str(spec)[-1] != "\n"
+
+
+def test_reload(specfile_factory, spec_minimal, spec_traditional):
+    spec = specfile_factory(spec_minimal)
+    before_reload = copy.deepcopy(spec)
+
+    if spec.path is not None:
+        os.remove(spec_minimal)
```
nforro, line 696: You could make this parametrized with e.g. `remove_spec` argument that would be first `False` and then `True`, to test both cases.

tests/integration/test_specfile.py
```diff
@@ -684,8 +681,35 @@ def test_update_version(
...
+        shutil.copy2(spec_traditional, spec_minimal)
+        spec.reload()
+        after_reload = spec
+    else:
+        spec._file.seek(0)
+        spec._file.truncate()
+        content = Path(spec_traditional).read_text(
```
nforro, line 703: `spec_traditional` is a fixture returning a `Path`, so you can definitely drop `Path()` here (sorry, I didn't realize that before).

specfile/specfile.py
nforro: Ok, if you want to keep the current approach, there are two things:
- nothing should happen if the path hasn't changed, i.e. if `value == self.path`
- if `self._file` is e.g. `StringIO`, its buffer is discarded when it is closed, so a user could lose modifications (though one could ask why would they be setting path in such case), so perhaps it's better not to explicitly close the old file (if there is no reference to it outside of `Specfile` instance it will be closed implicitly)

nforro (commented):
Very nice, just a couple of last remarks.

mynk8:
> I've (finally) noticed there are a lot of mypy errors. They all need to be addressed, I'll try to go through them and add suggestions.

Yes, there were mypy errors, mostly about `IOBase` not having attributes due to duck typing.

specfile/specfile.py
```diff
@@ -50,9 +65,13 @@ class Specfile:
         autosave: Whether to automatically save any changes made.
     """
 
+    ENCODING_ARGS = {"encoding": "utf8", "errors": "surrogateescape"}
```
nforro, line 68: To stop mypy from complaining about this, it is necessary to assign types to specific keys (`TypedDict` is from `typing`):

`EncodingArgs = TypedDict("EncodingArgs", {"encoding": str, "errors": str})`

And then use the type:

```suggestion
    ENCODING_ARGS: EncodingArgs = {"encoding": "utf8", "errors": "surrogateescape"}
```

specfile/specfile.py
```diff
@@ -71,12 +92,29 @@ def __init__(
...
+            self._file = file
+        elif path is not None:
+            self._file = Path(path).open("r+", **self.ENCODING_ARGS)
+        elif content is not None:
+            self._file = StringIO(content)
+        else:
+            raise ValueError("Either `file`, `path` or `content` must be provided")
+        if sourcedir is None:
+            try:
+                sourcedir = Path(self._file.name).parent
```
nforro, line 110: Here we are treating `self._file` as named file.

```suggestion
                sourcedir = Path(cast(FileIO, self._file).name).parent
```

specfile/specfile.py
```diff
@@ -111,6 +149,43 @@ def __exit__(
...
+        specfile = self.__class__.__new__(self.__class__)
+        memodict[id(self)] = specfile
+
+        for k, v in self.__dict__.items():
+            if k == "_file":
+                continue
+            setattr(specfile, k, copy.deepcopy(v, memodict))
+
+        try:
+            path = Path(self._file.name)
```
nforro, line 165: Here we are again treating `self._file` as named file.

```suggestion
            path = Path(cast(FileIO, self._file).name)
```

specfile/specfile.py
```diff
@@ -111,6 +149,43 @@ def __exit__(
...
+        except AttributeError:
+            # not a named file, try `getvalue()`
+            try:
+                specfile._file = type(self._file)(self._file.getvalue())
```
nforro, line 169: Here we are treating `self._file` as either `StringIO` or `BytesIO` (or their subclasses).

```suggestion
                specfile._file = type(cast(Union[StringIO, BytesIO], self._file))(
                    cast(Union[StringIO, BytesIO], self._file).getvalue()
                )
```

You could also define a new type for the union to make the code more compact.

specfile/specfile.py
```diff
@@ -111,6 +149,43 @@ def __exit__(
...
+                # and copy data
+                specfile._file = type(self._file)()
+                specfile._file.write(self._file.read())
+                specfile._file.seek(0)
+        else:
+            try:
+                specfile._file = path.open(
+                    mode=self._file.mode,
+                    encoding=self._file.encoding,
+                    errors=self._file.errors,
```
nforro, line 181: Here we are treating `self._file` as named file opened in text mode.

```suggestion
                    mode=cast(TextIOBase, self._file).mode,
                    encoding=cast(TextIOBase, self._file).encoding,
                    errors=cast(TextIOBase, self._file).errors,
```

specfile/specfile.py
```diff
@@ -111,6 +149,43 @@ def __exit__(
...
+                )
+            except AttributeError:
+                # files open in binary mode have no `encoding`/`errors`
+                specfile._file = path.open(mode=self._file.mode)
```
nforro, line 185: Here we are treating `self._file` as named file again.

```suggestion
                specfile._file = path.open(mode=cast(FileIO, self._file).mode)
```

specfile/specfile.py
```diff
@@ -119,19 +194,34 @@ def _dump_debug_info(self, message) -> None:
...
+            content = raw_content.decode(**cls.ENCODING_ARGS)
+        return content.splitlines(), content.endswith("\n")
 
     @property
-    def path(self) -> Path:
+    def path(self) -> Optional[Path]:
         """Path to the spec file."""
-        return self._path
+        try:
+            return Path(self._file.name)
```
nforro, line 211: Here we are treating `self._file` as named file again.

```suggestion
            return Path(cast(FileIO, self._file).name)
```

specfile/specfile.py
```diff
@@ -179,11 +269,26 @@ def rpm_spec(self) -> rpm.spec:
 
     def reload(self) -> None:
         """Reloads the spec file content."""
-        self._lines, self._trailing_newline = self._read_lines(self.path)
+        try:
+            path = Path(self._file.name)
```
nforro, line 273: Here we are treating `self._file` as named file again.

```suggestion
            path = Path(cast(FileIO, self._file).name)
```

nforro (commented):
Ok, here are suggestions for the mypy errors.

nforro (commented):
Some more fixes, however I found out that `typing` provides more IO types, maybe it would be better to use those.

specfile/specfile.py
```diff
@@ -111,6 +152,45 @@ def __exit__(
...
+                specfile._file = type(self._file)()
+                specfile._file.write(self._file.read())
+                specfile._file.seek(0)
+        else:
+            try:
+                specfile._file = path.open(
+                    mode=cast(TextIOBase, self._file).mode,
+                    encoding=cast(TextIOBase, self._file).encoding,
+                    errors=cast(TextIOBase, self._file).errors,
+                )
```
mynk8, line 187:
```suggestion
                specfile._file = cast(TextIOWrapper,
                    path.open(
                    mode=cast(TextIOBase, self._file).mode,
                    encoding=cast(TextIOBase, self._file).encoding,
                    errors=cast(TextIOBase, self._file).errors,
                ))
```
This was needed to silence,
`Incompatible types in assignment (expression has type "IO[Any]", variable has type "IOBase")  [assignment] (185:33)`

specfile/specfile.py
```diff
@@ -50,9 +68,13 @@ class Specfile:
         autosave: Whether to automatically save any changes made.
     """
 
+    ENCODING_ARGS: EncodingArgs = {"encoding": "utf8", "errors": "surrogateescape"}
+
     def __init__(
         self,
-        path: Union[Path, str],
+        path: Optional[Union[Path, str]] = None,
+        file: Optional[IOBase] = None,
```
gotmax23, line 76: I think this should be annotated with `typing.IO[str]` (a stream opened in string mode) and not `IOBase`.

specfile/specfile.py
```diff
@@ -63,6 +85,8 @@ def __init__(
 
         Args:
             path: Path to the spec file.
+            file: File object representing the spec file.
```
gotmax23, line 88: Would it be better to name this `stream`? `file` doesn't clearly communicate that it's supposed to be an open file stream, not a filename/path.

specfile/specfile.py
gotmax23: Or `typing.IO[str] | typing.IO[bytes]`, since the code also works with files opened in binary mode.

gotmax23:
https://gist.github.com/gotmax23/9904de7f2d581e51da9ef4ae7e2d5030 should fix the typing issues. I just annotated everything using plain `IO`, as it made typing a bit easier. The type checker really didn't like the code in https://gist.github.com/gotmax23/9904de7f2d581e51da9ef4ae7e2d5030#file-0001-specfile-fix-typing-of-file-streams-patch-L89 (is it even necessary in the first place, though?) when I typed this as `IO[str] | IO[bytes]`. It also might make sense to only accept file streams opened in text mode. Trying to deal with both makes this start to smell like Python 2 compat code...

specfile/specfile.py
mynk8: I think it was meant to represent a file object as well

mynk8:
:thinking: I don't understand why shouldn't we ? I feel it generalises it for use with more types.

nforro:
> https://gist.github.com/gotmax23/9904de7f2d581e51da9ef4ae7e2d5030 should fix the typing issues. I just annotated everything using plain `IO`, as it made typing a bit easier.

Thanks, I was going to suggest using `IO` and `TextIO` from `typing` as well.

> The type checker really didn't like the code in https://gist.github.com/gotmax23/9904de7f2d581e51da9ef4ae7e2d5030#file-0001-specfile-fix-typing-of-file-streams-patch-L89 (is it even necessary in the first place, though?) when I typed this as `IO[str] | IO[bytes]`.

Right, we should probably just drop that code.

> It also might make sense to only accept file streams opened in text mode. Trying to deal with both makes this start to smell like Python 2 compat code...

I don't think accepting also binary streams complicates the code all that much and it could be useful IMHO.

specfile/specfile.py
nforro: Python glossary lists [file object](https://docs.python.org/3/glossary.html#term-file-object) and [file-like object](https://docs.python.org/3/glossary.html#term-file-like-object), but `stream` is also mentioned as a synonym.

specfile/types.py
```diff
@@ -4,11 +4,14 @@
 import abc
 
 try:
-    from typing import SupportsIndex
+    from typing import SupportsIndex, TypedDict
```
nforro, line 7: This shouldn't be here, although it's true that `TypedDict` doesn't exist in Python 3.6 either. Still, I wouldn't couple these two typed together and rather use a separate `try`-`except` for `TypedDict`.

specfile/specfile.py
```diff
@@ -50,9 +65,14 @@ class Specfile:
         autosave: Whether to automatically save any changes made.
     """
 
+    ENCODING_ARGS: EncodingArgs = {"encoding": "utf8", "errors": "surrogateescape"}
+    _file: IO
```
nforro, line 69: Is this really necessary? Isn't the type inferred from the first assignment to `self._file`?

specfile/specfile.py
```diff
@@ -111,6 +149,39 @@ def __exit__(
...
+            setattr(specfile, k, copy.deepcopy(v, memodict))
+
+        try:
+            path = Path(self._file.name)
+        except AttributeError:
+            # IO doesn't implement getvalue() so tell mypy this is StringIO
+            # (could also be BytesIO)
+            sio = cast(StringIO, self._file)
+            # not a named file, try `getvalue()`
+            specfile._file = type(sio)(sio.getvalue())
```
nforro, line 171: I'm not a fan of this, but I don't know how to do this better without actually checking for type.

gotmax23: You could also a type ignore, but I think that's worse.

mynk8: I can't think of any alternative method so far.

specfile/specfile.py
```diff
@@ -63,6 +82,8 @@ def __init__(
 
         Args:
             path: Path to the spec file.
+            file: File object representing the spec file.
+            content: String containing the spec file content.
```
nforro, line 86: Can we document the fact that these arguments are mutually exclusive? Probably in the description above, to avoid repetition.

specfile/types.py
```diff
@@ -12,3 +12,11 @@
...
         @abc.abstractmethod
         def __index__(self) -> int: ...
+
+
+try:
+    from typing import TypedDict
+except ImportError:
+    from typing_extensions import TypedDict
+
+EncodingArgs = TypedDict("EncodingArgs", {"encoding": str, "errors": str})
```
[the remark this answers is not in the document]
nforro: This shows as resolved, but nothing has changed.

nforro (approved):
Thank you for the work on this!
Could you please squash the commits and update the release notes?

specfile/specfile.py
nforro: Also, maybe it would make sense to change the order and swap `file` and `content`, `file` is on a lower level than `path` and `content` and we can't move `path` for backward compatibility.

specfile/types.py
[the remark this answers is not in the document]
mynk8: I think i forgot to commit that change my bad

specfile/specfile.py
mynk8:
```python
            specfile._file = StringIO(cast(StringIO, self._file).getvalue())
```
I think this could be an alternative way to write it.

mynk8:
Thanks for actively guiding throughout the process !!
I learnt a lot working on this.

specfile/specfile.py
```diff
@@ -50,19 +65,25 @@ class Specfile:
...
+        file: Optional[IO] = None,
         sourcedir: Optional[Union[Path, str]] = None,
         autosave: bool = False,
         macros: Optional[List[Tuple[str, Optional[str]]]] = None,
         force_parse: bool = False,
     ) -> None:
         """
         Initializes a specfile object.
-
+        The arguments `path`, `file` and `content` are mutually exclusive and only one can be used.
```
nforro, line 82: I'm not able to apply the suggestion from the start of the docstring, nevertheless:

```suggestion
        Initializes a specfile object. You can specify either a path to the spec file,
        its content as a string or a file object representing it. `sourcedir` is optional
        if `path` or a named `file` is provided and will be set to the parent directory.
        
```

nforro:
Could you modify the description a bit:
> Fixes
> Related to https://github.com/packit/specfile/issues/206 and https://github.com/packit/specfile/issues/248

It's not only related to those issues, it does fix them.

> RELEASE NOTES BEGIN
> 
> - Support creating Specfile instances from file-objects and strings.
> - reload() now reopens the file.
> - Update tests for running every test case with a fixture for different input modes, file_path, text_file, binary_file, text_io_stream, > binary_io_stream, content_string.
> - Add a test for reload().
> 
> RELEASE NOTES END

Release notes are supposed to describe user-facing changes. `Added support for creating Specfile instances from file objects and strings.` is just fine.

specfile/specfile.py
```diff
@@ -50,19 +65,26 @@ class Specfile:
...
         autosave: bool = False,
         macros: Optional[List[Tuple[str, Optional[str]]]] = None,
         force_parse: bool = False,
     ) -> None:
         """
-        Initializes a specfile object.
-
+        Initializes a specfile object. You can specify either a path to the spec file,
+        its content as a string or a file object representing it. `sourcedir` is optional
+        if `path` or a named `file` is provided and will be set to the parent directory.
```
nforro, line 83: Just the missing newline here: https://github.com/packit/contributing?tab=readme-ov-file#docstrings

```suggestion
        if `path` or a named `file` is provided and will be set to the parent directory.
        
```

nforro:
> Fixes #206 and #248

Actually, in order for GitHub to automatically close the fixed issues you have to list them separately, so one `Fixes` line per each. Note the `Successfully merging this pull request may close these issues.` section on the right.

PR #463, opened 2025-03-20
ci(pypi): split the PyPI release process

mfocko:
Related to packit/private#32

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -18,14 +18,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
+          fetch-tags: true
           persist-credentials: false
 
       - uses: actions/setup-python@v5
-
-      - name: Get history and tags for SCM versioning to work
-        run: |
-          git fetch --prune --unshallow
-          git fetch --depth=1 origin +refs/tags/*:refs/tags/*
 
       - name: Build a source tarball and a binary wheel
         # https://pypa-build.readthedocs.io
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -5,23 +5,57 @@ name: Publish Python 🐍 distributions 📦 to PyPI
 
 on:
   release:
+    # • `released` reacts to stable releases
+    # • `prereleased` reacts to pre-releases
+    # • `published` reacts to both
     types: [published]
 
 jobs:
-  build-n-publish:
+  build:
+    name: Build distribution 📦
     runs-on: ubuntu-latest
-    permissions:
-      id-token: write # for trusted publishing
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/pypi-publish.yml
```diff
@@ -5,26 +5,57 @@ name: Publish Python 🐍 distributions 📦 to PyPI
 
 on:
   release:
-    types: [published]
+    types: [released]
```
mfocko, line 8: What's the difference here? :eyes:

nforro: That's for prereleases: https://github.com/packit/specfile/commit/babf6b0c41459b5a4c3d56eabec631c0693665c6

.github/workflows/pypi-publish.yml
```diff
@@ -5,24 +5,58 @@ name: Publish Python 🐍 distributions 📦 to PyPI
...
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - uses: actions/setup-python@v5
+
+      - name: Get history and tags for SCM versioning to work
+        run: |
+          git fetch --prune --unshallow
+          git fetch --depth=1 origin +refs/tags/*:refs/tags/*
```
nforro, line 28: How come it worked until now without this?

nforro (approved):

.github/workflows/pypi-publish.yml
mfocko: have no clue, maybe it was a change in the way github action checkouts the source? I could have a look how we got to this point, but I also think that we started without this and it got added at some point

gotmax23, line 28: Why not pass `fetch-depth` to the checkout action?

mfocko: https://github.com/packit/packit/commit/bf36c82d7fb511b3a8c573ff3278c350317edfa7

mfocko: weren't the scm tools dropped since then? :eyes: if it's related to setuptools_scm

mfocko: I was actually checking the options they have, but I was not able to combine it with the first `fetch --prune --unshallow`

gotmax23: What's the purpose of that command? If you simply want to do a full clone so that setuptools_scm works properly, you can set fetch-depth to 0. Or is there something else here?

gotmax23: setuptools_scm is still around (https://github.com/packit/specfile/blob/875f03eda8abb51e7211bd1de5c11bdb038cf03d/setup.py#L15, https://github.com/packit/specfile/blob/875f03eda8abb51e7211bd1de5c11bdb038cf03d/pyproject.toml#L9).

PR #470, opened 2025-05-15
Reopen named files also before saving

nforro:
The underlying file could have been deleted or replaced and its inode may be different, reopening it just before writing should ensure the content is written where it's supposed to.


mfocko (approved):

lbarcziova (approved):
thanks!

PR #482, opened 2025-06-15
Add author and timestamp properties to ChangelogEntry

dcermak:
These properties are required to convert Fedora to openSUSE style changelogs for packit

TODO:

- [-] Update or write new documentation in `packit/packit.dev`.

Fixes

Related to

Merge before/after

RELEASE NOTES BEGIN
- add support to parse the author and timestamp from changelog entries
RELEASE NOTES END

### DIFF
--- a/specfile/changelog.py
+++ b/specfile/changelog.py
@@ -194,6 +194,77 @@ class ChangelogEntry:
             return ChangelogStyle.openSUSE
         return ChangelogStyle.standard
 
+    @property
+    def author(self) -> Optional[str]:
+        """The author of this changelog entry including their email."""
+        if self.style == ChangelogStyle.openSUSE:
+            # openSUSE format: "OPENSUSE_CHANGELOG_SEPARATOR\n$DATE - $AUTHOR <$EMAIL>"
+            # Extract everything after " - " from the actual header line (second line)
+            header_lines = self.header.split("\n")
+            if len(header_lines) > 1:
+                if len(date_author := header_lines[1].split("-", maxsplit=1)) > 1:
+                    return date_author[1].strip()
+        else:
+            # Standard format: "* $DATE $AUTHOR <$EMAIL> [- $EVR]"
+            header_without_asterisk = self.header.removeprefix("* ")
+
+            # Find the end of the date portion by looking for year (4 digits followed by space)
+            year_match = re.search(r"\d{4}\s+", header_without_asterisk)
+            if year_match:
+                author_and_evr = header_without_asterisk[year_match.end() :]
+
+                if not self.evr:
+                    return author_and_evr
+
+                if (gt_ind := author_and_evr.rfind(">")) > 0:
+                    return author_and_evr[: gt_ind + 1].strip()
+
+        return None
+
+    @property
+    def timestamp(self) -> datetime.datetime:
+        """The timestamp of this changelog entry."""
+        if self.style == ChangelogStyle.openSUSE:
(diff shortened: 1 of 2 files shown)


specfile/changelog.py
```diff
@@ -194,6 +194,77 @@ def style(self) -> ChangelogStyle:
...
+            # Find the end of the date portion by looking for year (4 digits followed by space)
+            year_match = re.search(r"\d{4}\s+", header_without_asterisk)
+            if year_match:
+                author_and_evr = header_without_asterisk[year_match.end() :]
+
+                if not self.evr:
+                    return author_and_evr
+
+                if (gt_ind := author_and_evr.rfind(">")) > 0:
+                    return author_and_evr[: gt_ind + 1].strip()
```
nforro, line 220: Wouldn't it be possible to use `_EVR_RE` for this?

specfile/changelog.py
```diff
@@ -194,6 +194,77 @@ def style(self) -> ChangelogStyle:
...
+
+            # Find the end of the date portion by looking for year (4 digits followed by space)
+            year_match = re.search(r"\d{4}\s+", header_without_asterisk)
+            if year_match:
+                author_and_evr = header_without_asterisk[year_match.end() :]
+
+                if not self.evr:
+                    return author_and_evr
+
+                if (gt_ind := author_and_evr.rfind(">")) > 0:
```
nforro, line 219: This assumes the author substring always ends with a specifically formatted e-mail, but that doesn't have to be the case, even though Fedora Packaging Guidelines [mandate](https://docs.fedoraproject.org/en-US/packaging-guidelines/manual-changelog/) so. specfile should be able to parse any format supported by RPM, which allows any string after the timestamp (ended by a newline) - and refers to it as `name`. What about adding a more generic `name` property and then deriving `author` and `evr` from it?

specfile/changelog.py
```diff
@@ -194,6 +194,77 @@ def style(self) -> ChangelogStyle:
...
+                    return author_and_evr[: gt_ind + 1].strip()
+
+        return None
+
+    @property
+    def timestamp(self) -> datetime.datetime:
```
nforro, line 225: Could this return `datetime.date` for non-extended timestamps, to mirror the behavior of `add_changelog_entry()`?

nforro (commented):
> To properly parse the timezone, we'd have to add another dependency. What's your opinion on this? Add a dependency or live with wrong timezones?

That's one of the reasons this was never implemented. Wrong timezones are a no-go, if we are going to implement this, let's do it right.

nforro:
> To properly parse the timezone, we'd have to add another dependency

Perhaps adding another dependency could be avoided by matching the entries to RPM-parsed changelog - `self.rpm_spec.sourceHeader[rpm.RPMTAG_CHANGELOGTIME]` returns a list of Unix timestamps.

specfile/changelog.py
nforro: That's true, but there are lots of spec files with changelog entries like this, for example:
https://src.fedoraproject.org/rpms/gdb/blob/rawhide/f/gdb.spec#_1516
https://src.fedoraproject.org/rpms/rpm-spec-language-server/blob/rawhide/f/rpm-spec-language-server.spec#_72

We could try to recognize these formats, and document that author/EVR parsing is best effort and doesn't have to be accurate:
- `* timestamp author - EVR`
- `* timestamp author [EVR]`
- `* timestamp author (EVR)`
- `* timestamp author`

PR #488, opened 2025-09-07
Remove `preserve_macros` parameter from https://github.com/packit/specfile#version-and-release

jessp01:
As far as I can see, that argument was never supported.


nforro (approved):
Good catch, thanks!

nforro:
regate

Issue #491, opened 2025-09-15
add_changelog_entry(): Incorrect `evr` expansion


jessp01:
### What happened? What is the problem?

For this specfile: https://gitlab.com/CentOS/archives/git.centos.org/rpms/openssh/-/blob/c7/SPECS/openssh.spec

The below code results in the wrong `evr` being set in the last changelog entry:
```python
#!/usr/bin/env python3
import sys

from specfile import Specfile

specfile = Specfile(sys.argv[1], force_parse=True)
changelog_entry = "This is a test changelog entry"
with specfile as spec:
    spec.bump_release()
    spec.add_changelog_entry(changelog_entry)
specfile.save()

```

$ git diff
```diff
@@ -71,7 +71,7 @@
 Summary: An open source implementation of SSH protocol versions 1 and 2
 Name: openssh
 Version: %{openssh_ver}
-Release: %{openssh_rel}%{?dist}%{?rescue_rel}
+Release: %{openssh_rel}%{?dist}%{?rescue_rel}.2
 URL: http://www.openssh.com/portable.html
 #URL1: http://pamsshagentauth.sourceforge.net
 Source0: ftp://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/openssh-%{version}.tar.gz
@@ -842,6 +842,9 @@ getent passwd sshd >/dev/null || \
 %endif
 
 %changelog
+* Mon Sep 15 2025 Jesse Portnoy <jesse.portnoy@perforce.com> - 0.10.3-2.23
+This is a test changelog entry
+
 * Thu Jul 20 2023 Dmitry Belyavskiy <dbelyavs@redhat.com> - 7.4p1-23 + 0.10.3-2
 - Avoid remote code execution in ssh-agent PKCS#11 support
   Resolves: CVE-2023-38408
```

In other words, the `evr` is set to `0.10.3-2.23` (the value of `%{pam_ssh_agent_ver}-%{pam_ssh_agent_rel`)

### What did you expect to happen?

In other changelog entries, the version is set to `%{version}-%{release} + %{pam_ssh_agent_ver}-%{pam_ssh_agent_rel}`. Obviously, the two latter macros are very specific to this specfile and I have no expectation that we'll analyse all specfile macros in an attempt to find such things but I did expect the result to be the same as what I get with the below:

```python
with specfile as spec:
spec.add_changelog_entry(changelog_entry, evr=f"{spec.expanded_version}-{spec.expanded_release}")
```

To wit:
```diff
[...]

PR #492, opened 2025-09-16
Fixes: #491: correct add_changelog_entry(): Incorrect evr expansion 

jessp01:
Address https://github.com/packit/specfile/issues/491: add_changelog_entry(): Incorrect evr expansion

### DIFF
--- a/specfile/specfile.py
+++ b/specfile/specfile.py
@@ -611,7 +611,15 @@ class Specfile:
                 if self.contains_autochangelog(section):
                     continue
                 if evr is None:
-                    evr = "%{?epoch:%{epoch}:}%{version}-%{release}"
+                    with self.tags() as tags:
+                        try:
+                            evr = f"{tags.epoch.expanded_value}:"
+                        except AttributeError:
+                            evr = ""
+                        evr += f"{tags.version.expanded_value}-"
+                        evr += self.expand(
+                            tags.release.value, extra_macros=[("dist", "")]
+                        )
                 with self.changelog(section) as changelog:
                     if changelog is None:
                         return


nforro (approved):
Thanks. I wonder if we shouldn't disable using macros in `evr` at all, considering this issue (however rare).

nforro:
Could you please rebase and squash the commits into one?

specfile/specfile.py
```diff
@@ -611,7 +611,12 @@ def add_changelog_entry(
                 if self.contains_autochangelog(section):
                     continue
                 if evr is None:
-                    evr = "%{?epoch:%{epoch}:}%{version}-%{release}"
+                    epoch_prefix = (
+                        f"{self.expanded_epoch}:" if self.expanded_epoch else ""
```
gotmax23, line 615: Should this use the tags() context manager directly for better performance instead of two `expanded_*` properties that cause the tags context to be instantiated twice?

gotmax23: Also, the values are expanded a second time on 623 which is probably not desired.

specfile/specfile.py
```diff
@@ -611,7 +611,12 @@ def add_changelog_entry(
...
+                    )
+                    evr = (
+                        f"{epoch_prefix}{self.expanded_version}-{self.expanded_release}"
+                    )
```
nforro, line 619:
```suggestion
                    with self.tags() as tags:
                        try:
                            evr = f"{tags.epoch.expanded_value}:"
                        except AttributeError:
                            evr = ""
                        evr += f"{tags.version.expanded_value}-{tags.release.expanded_value}"
```

specfile/specfile.py
```diff
@@ -611,7 +611,12 @@ def add_changelog_entry(
                 if self.contains_autochangelog(section):
                     continue
                 if evr is None:
-                    evr = "%{?epoch:%{epoch}:}%{version}-%{release}"
+                    with self.tags() as tags:
+                        try:
+                            evr = f"{tags.epoch.expanded_value}:"
+                        except AttributeError:
+                            evr = ""
+                        evr += f"{tags.version.expanded_value}-{tags.release.expanded_value}"
```
nforro, line 619:
```suggestion
                    with self.tags() as tags:
                        try:
                            evr = f"{tags.epoch.expanded_value}:"
                        except AttributeError:
                            evr = ""
                        evr += {tags.version.expanded_value}
                        evr += self.expand(tags.release.value, extra_macros=[("dist", "")])
```

specfile/specfile.py
jessp01: Yeap, units are failing with this. Shall we revert?

specfile/specfile.py
jessp01: I believe we're good with:
```python
                    with self.tags() as tags:
                        try:
                            evr = f"{tags.epoch.expanded_value}:"
                        except AttributeError:
                            evr = ""
                        evr += f"{tags.version.expanded_value}-"
                        evr += self.expand(
                            tags.release.value, extra_macros=[("dist", "")]
                        )
```

Marking as resolved.

PR #532, opened 2026-04-23
Avoid opening Specfile with write-access

praiskup:
Per discussion in Copr issue tracker:
https://github.com/fedora-copr/copr/issues/4266

### DIFF
--- a/specfile/specfile.py
+++ b/specfile/specfile.py
@@ -106,7 +106,7 @@ class Specfile:
         if file is not None:
             self._file = file
         elif path is not None:
-            self._file = Path(path).open("r+", **self.ENCODING_ARGS)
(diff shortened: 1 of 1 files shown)


praiskup:
Seems this change is indeed wrong.  Do whatever you prefer with this PR (guide me to provide a better patch, but even closing is fine).  It was easier for me to open this than filling a proper issue...

nforro:
specfile is primarily read-write by design. I think the proper fix here would be to try to open the path with `r+`, falling back to `r` on `PermissionError` and preserving the mode during reopen (or perhaps trying again `r+` with `r` as fallback).

praiskup:
Ok, lemme close this one and have an issue instead: https://github.com/packit/specfile/issues/534

PR #537, opened 2026-05-20
Pin GitHub Actions to commit hashes for security

majamassarini:
Pin github actions to specific commit SHAs to prevent supply chain attacks and ensure reproducible builds.

Assisted-By: Claude Sonnet 4.5

### DIFF
--- a/.github/workflows/check-release-notes.yml
+++ b/.github/workflows/check-release-notes.yml
@@ -10,6 +10,6 @@ on:
 jobs:
   check_release_notes:
     name: Notes are either written, or there are none
-    uses: packit/.github/.github/workflows/check-release-notes.yml@main
+    uses: packit/.github/.github/workflows/check-release-notes.yml@2837c96caf71966609451ad0323552ef4be11a23 # main
     with:
       description: ${{ github.event.pull_request.body }}
--- a/.github/workflows/do-release.yml
+++ b/.github/workflows/do-release.yml
@@ -11,7 +11,7 @@ jobs:
     if: github.event.pull_request.merged == true && contains(github.event.pull_request.labels.*.name, 'release') && github.repository_owner == 'packit'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3
       - name: Create GitHub release
         run: |
           VERSION=$(grep -oP '^# \K[0-9.]+([.\-_]?[a-zA-Z0-9.\-_]+)?' CHANGELOG.md | head -n 1)
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -16,18 +16,18 @@ jobs:
       contents: write
 
     steps:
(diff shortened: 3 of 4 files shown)


nforro:
> Pin all non-packit actions in artifact-handling workflows to specific commit SHAs to prevent supply chain attacks and ensure reproducible builds.

Why exclude Packit actions?

majamassarini:
> > Pin all non-packit actions in artifact-handling workflows to specific commit SHAs to prevent supply chain attacks and ensure reproducible builds.
> 
> Why exclude Packit actions?

because I think we trust our team/process more the external ones and I suppose the risk is therefore lower with our own actions. But if you think it is better to pin down all of them, then I will do it.

nforro:
> because I think we trust our team/process more the external ones and I suppose the risk is therefore lower with our own actions

I don't think this is about trust, do you think it's less likely for an attacker to get access to one of our repos and "retarget" a tag than to any third-party repo?

majamassarini:
> > because I think we trust our team/process more the external ones and I suppose the risk is therefore lower with our own actions
> 
> I don't think this is about trust, do you think it's less likely for an attacker to get access to one of our repos and "retarget" a tag than to any third-party repo?

Not less likely than for a GitHub repo, but for other third-party repos yes, I consider it less likely. However, I see your point, and I will change our actions too.

nforro (approved):

.github/workflows/check-release-notes.yml
```diff
@@ -10,6 +10,6 @@ on:
 jobs:
   check_release_notes:
     name: Notes are either written, or there are none
-    uses: packit/.github/.github/workflows/check-release-notes.yml@main
+    uses: packit/.github/.github/workflows/check-release-notes.yml@2837c96caf71966609451ad0323552ef4be11a23 # main
```
LecrisUT, line 13: This has quite a trade-off of needing to babysit the these updates based on commit bumps to main. May not be that big of an issue if everyone remembers to bump when the relevant files in `packit/packit` change.

.github/workflows/do-release.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     if: github.event.pull_request.merged == true && contains(github.event.pull_request.labels.*.name, 'release') && github.repository_owner == 'packit'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3
```
LecrisUT, line 14: Note that a per-requisite for these is to change the tag from `vX`  to `vX.Y.Z`, otherwise tools like dependabot and renovate would not work properly (yes it would just freeze to the commit you gave, but also not give you any updates either). Although, these have not been updated since `v3` so far anyway 😅.

majamassarini: Thanks @LecrisUT for the tip, I didn't know that. However as you noticed, we haven't updated them so far. When I was pinning these down, I was wondering if it was a good moment to update them all. But I dismissed it, because there is no easy way to test that I don't break anything. So I think we can keep living with these versions. 😅

